PR #4, opened 2023-11-17
add cache_mode and draft_model_dir to config_sample.yml

waldfee:
Had to study the code to find out how to set 8bit cache and speculative sampling. I think it makes sense to have it in the config sample.


bdashore3:
I'm going to merge this PR even though there's no API support. API additions require a bit more work and the support from config.yml works for now.

PR #11, opened 2023-11-19
Fix incorrect ratio calculation for draft model

veden:


bdashore3:
@veden Hi there! Thanks for contributing.

For us to look at this PR, please include a description of the PR and a rationale of why this change should be implemented. That will help us understand your reasoning. Thanks!

Please look at [contributions](https://github.com/theroyallab/tabbyAPI#contributing) for reference.

turboderp:
Thanks. This was an oversight. Apparently same oversight in ExUI as well, so double fix. :)

PR #13, opened 2023-11-28
Adding the usage stat support (prompt_tokens, completion_tokens, and total_tokens)

ziadloo:
Out of the 4 methods of using the API (completion, completion+streaming, chat, chat+streaming), the usage stat was not added to streaming version of the chat. That's because unlike the other 3, this one's response class did not have the usage stat's place holder. So I was not sure if that response is supposed to have it. Adding the support to that response is fairly easy, if needed.


bdashore3:
There's still a small problem with the EOS token not being counted in the final UsageStats output, but we can ignore this for now and come back to it later. Merging.

PR #26, opened 2023-12-09
Reverting to use as the gpu_split_auto also evaluates to true

slaser79:
It appears that gpu_split_auto always evaluates to True even if set to False due to `or True` defaulting when the gpu_split_auto has been set to False.


bdashore3:
I did not notice this at all somehow. Good catch!

Fully fixed in 5ae2a91c04a437b369be193628c19d4c03edd9ef

PR #27, opened 2023-12-10
Loading Conversation Template based on Model Name and Fix DockerCompose

marcasmed:
## Conversation Template

The `get_chat_completion_prompt` function in the `utils` module currently takes the name of the loaded model as an argument and uses it to generates a conversation object from `fastchat.model.model_adapter`. Regardless of the model's name, the object returned is a `one_shot` conversation object. I suspect this is for a future implementation.

Since I use and test different Hugging Face models, I am working on an interface to load the correct fastchat conversation object. Currently, it only detects Chat-ML based templates, as these are the templates I use the most. However, extending it to other potential models should be easy.

The detection algorithm is based on the text-generation-ui implementation for this task, so credit goes to oobabooga.

## Dockerfile
Currently, the Dockerfile copies the `config_sample.yml` file in the root directory and renames it to config.yml regardless of whether this file already exists. This edit fixes that by simply deleting this line (as the app only needs the `config.yml` file if personalizations are made)

As well, it moves the `COPY . .`  command to avoid unnecessary downloads and re-installations.


marcasmed:
Update: 

Added support for:
- Mistral
- Alpaca
- Llama-v2

bdashore3:
I'm a bit confused on what's supposed to be done here. Currently, fastchat is only used as a shim to convert OAI chat completion lists into a prompt that Tabby can process. From my testing, all instruct formats (Alpaca, Vicuna, ChatML) get outputted into the proper prompt format for processing.

Ideally, I'd like to reduce the dependency on fastchat (possibly with in-house templates) in the future since the fschat package requires heavier dependencies such as Transformers and Accelerate. This dependency bloat is why fastchat is an optional dependency for those that need it.

Can you please elaborate more on why these changes are relevant in TabbyAPI?

marcasmed:
Agreed, the FastChat dependency could be removed with in-house templates. Currently, regardless of the name of the model, a message input of type list parses to a `one_shot` format conversation object. It is compatible, and the model will do inference, however the inference quality significantly drops as many models are not trained with that format.

## Example
Take the model [OpenHermes 2.5 - Mistral 7B](https://huggingface.co/LoneStriker/OpenHermes-2.5-Mistral-7B-8.0bpw-h8-exl2). If we were to mount it and send a post with a message list request to `/v1/chat/completions` :

```python
# main.py
@app.post("/v1/chat/completions", dependencies=[Depends(check_api_key), Depends(_check_model_container)])
async def generate_chat_completion(request: Request, data: ChatCompletionRequest):
   ...
   prompt = get_chat_completion_prompt(model_path.name, data.messages)
   ...

# utils
...
from fastchat.model.model_adapter import get_conversation_template
...

def get_chat_completion_prompt(model_path: str, messages: List[ChatCompletionMessage]):
   ...
   conv = get_conversation_template(model_path)


# fastchat.model.model_adapter.py
def get_conversation_template(model_path: str) -> Conversation:
    """Get the default conversation template."""
    adapter = get_model_adapter(model_path)
    return adapter.get_default_conv_template(model_path) # Fast chat is supposed to use the model name to obtain the default template for the model (see below)

...
def get_default_conv_template(self, model_path: str) -> Conversation:
        return get_conv_template("one_shot")    # The default template for all models, apparently is a 'one_shot' template...
```

Using the `one_shot` template will allow for inference; however the quality of the inference can drop significantly if it is not the format the model was trained on (and most models are not trained on that prompting style nowedays...)


## Proposal

A possible solution, would be to use the `get_conv_template` function directly from Tabby, which would need the user to pass, a correct Conversation template name:

```
# available names:
'raw', 'one_shot', 'zero_shot', 'vicuna_v1.1', 'airoboros_v1', 'airoboros_v2', 'airoboros_v3', 'koala_v1', 'alpaca', 'chatglm', 'chatglm2', 'chatglm3', 'codegeex', 'dolly_v2', 'oasst_pythia', 'oasst_llama', 'openchat_3.5', 'deepseek', 'tulu', 'stablelm', 'baize', 'rwkv', 'openbuddy', 'phoenix', 'ReaLM-7b-v1', 'chatgpt', 'claude', 'mpt-7b-chat', 'mpt-30b-chat', 'lemur-70b-chat', 'mpt-30b-instruct', 'bard', 'billa', 'redpajama-incite', 'h2ogpt', 'Robin', 'snoozy', 'manticore', 'falcon', 'polyglot_changgpt', 'tigerbot', 'xgen', 'internlm-chat', 'starchat', 'baichuan-chat', 'baichuan2-chat', 'mistral', 'llama-2', 'cutegpt', 'open-orca', 'mistral-7b-openorca', 'qwen-7b-chat', 'aquila-chat', 'aquila-legacy', 'aquila', 'aquila-v1', 'llama2-chinese', 'vigogne_instruct', 'vigogne_chat_v2', 'vigogne_chat_v3', 'falcon-chat', 'phind', 'metharme', 'zephyr', 'orca-2'
```

Since most of the models in Huggin Face have much different names:

```
LoneStriker_OpenHermes-2.5-Mistral-7B-8.0bpw-h8-exl2
LoneStriker_openinstruct-mistral-7b-6.0bpw-h6-exl2
TheBloke_Dolphin-Llama-13B-GPTQ
```

These names would need to either be manually mapped to an appropriate fast chat template name by the user (or the in-house preferred template once available). OR we would need a function capable to map these names to a FastChat conversation object.

My temporary solution uses the model name, to map to a set of templates and then from that set of templates, map to a FastChat conversation object.
- It memoizes the correct conversation object name.
- Registers a FastChat conversation object if it doesn't exist (this last part only implemented for Chat-ML based templates).

marcasmed:
Update: eliminated fast chat dependency

Refactored this modification to avoid having to import the Fast-Chat package.
Since it is largely based on their implementation, I added an acknowledgement section and their license.

bdashore3:
So, there are a few things I want to address here. Firstly, FastChat doesn't always return the one_shot template. It checks the model path name and returns an adapter based on what's provided. If an adapter can't be found, the base adapter is used.
(See [get_model_adapter](https://github.com/lm-sys/FastChat/blob/5aff3512c9d54821a5d183bb1ec234db4db13739/fastchat/model/model_adapter.py#L124) and [an example adapter](https://github.com/lm-sys/FastChat/blob/5aff3512c9d54821a5d183bb1ec234db4db13739/fastchat/model/model_adapter.py#L753))

I also want to reiterate that FastChat is optional. Removing it will involve a completely different system that's agnostic from FastChat's codebase to prevent bloating of TabbyAPI's code.

However, I did like your idea of adding an optional param to config and requests that specifies the prompt template of the model, which is much more lightweight and doesn't increase reliance on FastChat as a dependency.

As such, I have implemented that support in db87efde4aaeab0b5f3ae96d13c8430950e7b1ac. This change should be lightweight enough for users that want to specify their own template if they want to use chat completions with a list of message objects. 

But I highly recommend using a frontend that supports instruct out of the box (such as SillyTavern) so a formatted prompt in any instruct format can be sent to the API and parsed that way (via /v1/completions or /v1/chat/completions with a string).

If you'd still like to make changes to the docker-compose file. Please open an independent PR.

PR #31, opened 2023-12-19
cuda -> 12, pydantic error fixed.

veryamazinglystupid:
:3


DocShotgun:
The better fix would be just updating requirements.txt to specify a minimum version for pydantic, which now requires >= 2 (instead of version 1 when FastChat was being used).

PR #40, opened 2023-12-23
[WIP] setup.py and basic re-structuring

AlpinDale:
This PR lightly re-structures the repository by moving the main modules to the `tabbyapi` directory. It also adds a setup.py and pyproject.toml for creating wheels/releases.

The setup.py automatically recognizes the GPU on the device and installs the appropriate requirements file - if no GPU is found (or cuda/hip is not installed), it'll fall back to the cpu file.


Ph0rk0z:
Am enjoying not having any kind of setup, let alone installing this like a wheel. I just run start.sh and use my own environment that is shared.

AlpinDale:
@Ph0rk0z the start scripts won't be going anywhere. The plan here is to have a simple executable binary `tabbyapi` that can be called from anywhere as long as you're in the correct env. That should make it even easier than scripts.

Ph0rk0z:
That's true, it would still work that way. Mainly it's about the ease of doing git pulls and not having to do setup.

bsima:
Would love to have this merged. Having a proper build means I can use it on my nix system as a module/systemd service...

AlpinDale:
There's been a lot of internal changes here since I made this PR, so we need to remake this.

bdashore3:
@bsima you should already be able to use tabby in systemd by creating a service file and copying it over to `/etc/systemd/system`. At the moment, tabby has solidified itself as rolling release so proper versioning probably won't be a thing in the future.

I can provide an example `.service` file in the repo which should adhere to your usecase.

bsima:
I actually wrote a nix build for it this weekend that uses a modified pyproject.toml from the one on master. I want to get it tested and cleaned up a bit this week but then I'll share it here.

bdashore3:
Pyproject is already implemented. Closing this PR.

PR #54, opened 2024-01-24
New API overrides and formatting changes

bdashore3:
Changelog:
- Reformat the repository to make it readable and compartmentalized
- Add server-side sampler overriding (via YAML and API endpoints)
- Add ability to switch and unload templates via the API
- Add fasttensors support and expose generate_window


backends/exllamav2/model.py
```diff
@@ -163,30 +163,27 @@ def progress(loaded_modules: int, total_modules: int,
...
+                self.prompt_template = get_template_from_file(prompt_template_name)
+            except FileNotFoundError:
+                self.prompt_template = None
 
-            # Try finding the chat template from the model's config.json
-            # TODO: This may not even be used with huggingface models,
-            # mark for removal.
-            if self.prompt_template is None:
+            # Then try finding the template from the tokenizer_config.json
+            try:
```
djmaze, line 172: This implementation means I need to set a (dummy) `prompt_template` in the config even if it should use the template from tokenizer_config.json.

This is probably just wrong indendation?

djmaze: Oh, and even that does not work because it gets overridden in the next `try` block below that.

bdashore3: Fixed in the latest commit.

PR #78, opened 2024-03-07
Add Hugging Face repo download endpoint

DocShotgun:
Adds a wrapper for huggingface_hub downloader to download models/loras via API request, requires optional dependency huggingface_hub.

TODO:
- [ ] More robust reporting of errors that occur within snapshot_download such as a bad repo_id or revision, unauthorized access to a gated/private repo (currently results in a simple 500 Internal Server Error)
- [ ] Consider addition of default HF token to be used for downloading to config.yml, or reading a saved HF token from HF cache

N/A

Some users have requested the ability to download models through tabbyAPI itself. This adds a simple option that requires an optional dependency package for those who want this functionality and do not want to deal directly with huggingface-cli.

**Examples**
N/A

**Additional context**
N/A

### DIFF
--- a/OAI/types/download.py
+++ b/OAI/types/download.py
@@ -1,0 +1,12 @@
+""" Downloader types """
+from pydantic import BaseModel
+from typing import Optional
+
+
+class HFDownloadRequest(BaseModel):
+    """Represents a HuggingFace download request."""
+
+    repo_id: str
+    revision: Optional[str] = "main"
+    repo_type: Optional[str] = "model"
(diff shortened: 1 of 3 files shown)


bdashore3:
Great feature, but the current implementation of HF hub's downloader uses threads that don't have a SIGINT event set up. Therefore, this will require some more work to get working.

Reference: https://github.com/huggingface/huggingface_hub/issues/1818

bdashore3:
Implemented in 55ccd1b

PR #93, opened 2024-04-01
chore: make log level configurable via env variable

AlpinDale:
At the moment, you have to directly edit the logger file to change the level.


AlpinDale:
Ready for merge @bdashore3

PR #100, opened 2024-05-01
added new metrics to log_metrics: prompt_tokens_per_second

kir-gadjello:
Also shortened log names to fit in one line. Tested on my system.

### DIFF
--- a/backends/exllamav2/model.py
+++ b/backends/exllamav2/model.py
@@ -969,6 +969,7 @@ class ExllamaV2Container:
         generated_tokens = 0
         full_response = ""
         start_time = time.time()
+        first_token_time = None
         last_chunk_time = start_time
 
         save_tokens = torch.empty((ids.shape[0], 0), dtype=torch.bool)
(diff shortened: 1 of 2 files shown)


DocShotgun:
Took a look at the code - the problem is that this doesn't take into account the number of cached context tokens that didn't need to be re-processed.

If I prompt a generation with 2000 tokens of context, and then I prompt again with the same 2000 tokens plus an additional 500 tokens, the second generation will look like it has insanely fast prompt processing speed... because it didn't need to re-process the first 2000 tokens of the prompt.

kir-gadjello:
> Took a look at the code - the problem is that this doesn't take into account the number of cached context tokens that didn't need to be re-processed.
> 
> If I prompt a generation with 2000 tokens of context, and then I prompt again with the same 2000 tokens plus an additional 500 tokens, the second generation will look like it has insanely fast prompt processing speed... because it didn't need to re-process the first 2000 tokens of the prompt.

Yes, I have experienced this as well but the benefit of eyeballing the prompt processing stat was larger than this shortcoming. Maybe we could add that missing piece? I didn't see a simple way to achieve this reliably then though.

bdashore3:
Resolved by #109 because the new generator logs this in payloads.

PR #104, opened 2024-05-12
Fix: Actually pass banned_strings to the generation call.

ycros:
`banned_strings` is never added to `begin_stream_args`, so it's never passed into the `begin_stream_ex` call, and thus doesn't do anything.

This PR just adds it in.

Edit: `decode_special_tokens` was missing as well


bdashore3:
Mistake from dependency migration. Thanks!

PR #114, opened 2024-05-28
Fixed docker-compose.yml bind mounts

HeiressOfTheDark:
Updated the models bind mount and added a example one for a config.yml mount

Currently if you spin up a docker deployment via the docker-compose it bind mounts ./models to /usr/src/app/models when the containers working directory is /app not /usr/src/app

This allows for users to spin up the container without issues and gives them a example of how they can expand it.


bdashore3:
Solved in current dockerfile

Issue #127, opened 2024-06-03
AssertionError: Filter excluded all tokens


cikkle:
**Describe the bug**
I'm using the json_schema feature to constrain completion output and while it seems to work for schemas with a top-level object, it fails when I try to use an array at the root. Upon sending a request nothing happens until I ctrl+c and get the exception further below.

**To Reproduce**
This is a simple test schema I'm using, it passes in online validators:

```
{
    "$schema": "http://json-schema.org/draft-07/schema#",
    "type": "array",
    "items": {
        "type": "string"
    }
}
```

I send it to the api with an empty prompt. The other parameters look like this:

{
        "max_tokens": 1000,
        "generate_window": 512,
        "token_healing": true,
        "temperature_last": true,
        "add_bos_token": true,
        "ban_eos_token": false,
        "json_schema": (schema goes here),
        "stream": true,
        "response_format": { "type": "json_object" },
        "prompt": "",
        "temperature": 0.7,
        "top_p": 0.9,
        "top_k": 20,
        "typical_p": 1,
        "top_a": 0,
        "tfs": 1,
        "repetition_penalty": 1.15,
        "repetition_penalty_range": 0,
        "num_beams": 1,
        "length_penalty": 1,
        "encoder_rep_pen": 1,
        "do_sample": true,
        "early_stopping": false,
        "mirostat_mode": 0,
        "mirostat_tau": 5,
        "mirostat_eta": 0.1
}

**Expected behavior**
I expected it to work without error and constrain the output to something simple like:

>[ "hello" ]

**Logs**

> INFO:     ExllamaV2 version: 0.1.3
> INFO:     Your API key is: 
> INFO:     Your admin key is: 
> INFO:     
> INFO:     If these keys get compromised, make sure to delete api_tokens.yml and restart the server. Have fun!
> INFO:     Generation logging is disabled
> WARNING:  An unsupported GPU is found in this configuration. Switching to compatibility mode. 
> WARNING:  This disables parallel batching and features that rely on it (ex. CFG). 
> WARNING:  To disable compatability mode, all GPUs must be ampere (30 series) or newer. AMD GPUs are not supported.
> INFO:     Attempting to load a prompt template if present.
> INFO:     Using template "default" for chat completions.
> INFO:     Loading model: /home/o0/ai/models/text/command-r-v01-35B-exl2
> INFO:     Loading with autosplit
[...]

PR #129, opened 2024-06-03
[fix] Allow multiple prefixes with JSON constraint

turboderp:
Fix for #127

Allows first character of JSON completion to be either "[" or "{", preventing "Filter excluded all tokens" error when the JSON schema defines a top-level array.

**Examples**
N/A

**Additional context**
Requires exllamav2>=0.1.4

### DIFF
--- a/backends/exllamav2/grammar.py
+++ b/backends/exllamav2/grammar.py
@@ -79,8 +79,11 @@ class ExLlamaV2Grammar:
 
             return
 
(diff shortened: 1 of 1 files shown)


cikkle:
I don't know if I might be nitpicking, but primitives like `"hello"`, `true`, and `34985734.345` would also be valid at the top-level, wouldn't they?

PR #130, opened 2024-06-04
WIP: Add 'model' argument to OAI API to load a new model on the fly

bartowski1182:
Currently, when using tools like open-webui that support OpenAI endpoints (https://github.com/open-webui/open-webui), if you want to switch out your loaded model, there isn't an option besides making your own separate request to tabbyAPI or reloading the endpoint

It would be nice to support the OpenAI-like 'model' argument, and if that's not the currently loaded model, attempt to load it before generating

I don't have much experience with fastapi so I may have done some silly stuff with the key checking, please feel free to use this as more of an inspiration for a change than the actual appropriate code to use, I did my best to make it "proper" but it's still a bit ugly since I couldn't figure out how to re-use other parts (optional depends/model loading)


DocShotgun:
Took a brief look at the code. Seems like this only allows for default parameters for any model loaded via this method?

bartowski1182:
@DocShotgun correct, but not on purpose, just wasn't yet sure the proper way to include it without making everything uglier and it wasn't a starting priority for me

I'll see if there's something quick I can change if this is a feature worth continuing

bartowski1182:
I wonder how the API reacts to being told to load a model if dynamic gen is happening, maybe if this handles it the same that's okay?

It's also an optional param, and will behave the same as before when nothing is passed

Only thing is I should maybe include a config line to disable this in case of people using a frontend that is passing a model now, and this change would break their current setup

DocShotgun:
I suppose it wouldn't be hard to pass an entire model load config dict alongside the chat completion request - but will any OAI-compat frontend that supports model selection actually do this? Since the format of the model load request is very specific to tabby itself.

Normally the chat completion endpoint will add the new generation job to the queue, and exllamav2's dynamic generator will automatically get to it when it gets to it. This seems to add the model swap logic to the start of the router for chat completion requests, so its possible that it could just break anything that's already in the generation queue or being inferred, but I'm not sure.

bartowski1182:
valid points all around, I don't currently have any thoughts about this but they should definitely be considered

because of the lack of loading config options, it may make more sense to combine this with some kind of config where you can specify default loading behaviour per-model if there aren't any given

for now i'm happy to just understand that that's a limitation of requiring a model to load

another bonus to this feature is that, if you wanted to unload your model temporarily (like ollama does), that might be nice too, so you can load it as soon as you want to start a chat completion

bdashore3:
The need for this feature kinda overlaps with the existing routes since the dedicated load/unload endpoints already take in all the config options.

However, I'd be happy to add this param, with a few more questions:
- How should permissions work with this request? Should all users allowed to unload and load models with this arg?
- What would happen to other generations when this request is passed?
- How would the load parameters be specified (if at all?)

The config.yml can exist as a fallback loading config for this purpose as it's the default used on startup.

These points were somewhat mentioned above, but I'd like some more thorough planning to properly guide the implementation of this feature so that it doesn't compromise security.

I believe this feature (if added) should be an in-between and not a replacement for the dedicated endpoints with a config flag to allow insecure model loading.

bartowski1182:
> * How should permissions work with this request? Should all users allowed to unload and load models with this arg?

I was using the assumption that only admins would be able to, since that's what the load endpoint is behind, but would be happy if it was allowed to be anyways. Ultimately feel like it's not my call to make and either works!

> * What would happen to other generations when this request is passed?

Great question, I plan to look into how load/unload handle active generation and design from there

> * How would the load parameters be specified (if at all?)

I'm not sure yet but again would be great if it was possible hand any/all parameters to the load, otherwise it's stuck with default which is likely to be sub-optimal

That said, I'm not positive if any openai endpoints pass load params alongside the generation param.. could be first? Backup is saving/loading default configs per-model but that seems like a can of worms

> I believe this feature (if added) should be an in-between and not a replacement for the dedicated endpoints with a config flag to allow insecure model loading.

I agree 100%, definitely an augmentation and not a replacement

I also feel it best to throw it behind a config-level flag so that it defaults to off, avoiding potential breakage from people who happened to be passing a valid model name (open webui for example) and relied on it being ignored

ColumbusAI:
@bartowski1182 I tried using tabbyAPI with openwebui and get the following error in openwebui. Have you encountered this?

image
image

bartowski1182:
Quick guess is that you're giving openwebui your normal access token and you need to give the admin one for it to load models

ColumbusAI:
sheesh you were spot on. Thanks for the help and your PR!

bdashore3:
@bartowski1182 please check the commits I've made. I made the PR cleaner by using tabby's utility methods and unifying the overall function.

One thing I'm wondering is to error if a non-admin tries to load the model. It seems appropriate here.

bartowski1182:
It's beautiful, I'll test it out tomorrow with my openwebui fork! Seems like it should be identical though, glad you found a way to implement it that works with your style! 

This with the model config is gonna be so awesome

ro99:
@bartowski1182 by a model config you mean something like what exists for llama-cpp-python server?

https://llama-cpp-python.readthedocs.io/en/latest/server/#configuration-and-multi-model-support

PR #136, opened 2024-06-12
Fix disabling flash attention in exl2 config

DocShotgun:
Paged mode was being correctly disabled for unsupported setups, however exllamav2 requires that flash attention also be manually disabled for unsupported configurations even when falling back to compatibility mode.

Resolves https://github.com/theroyallab/tabbyAPI/issues/135

**Examples**
N/A

**Additional context**
This has the side effect of disabling flash attention for users with `2.2.1 <= flash-attn version < 2.5.7` to avoid additional checks. This was felt to be acceptable as all setups that support `flash-attn >= 2.2.1` will also support `flash-attn >= 2.5.7` and users will be prompted to upgrade.

### DIFF
--- a/backends/exllamav2/model.py
+++ b/backends/exllamav2/model.py
@@ -213,6 +213,7 @@ class ExllamaV2Container:
             self.config.no_flash_attn = True
             self.paged = False
             self.max_batch_size = 1
+            torch.backends.cuda.enable_flash_sdp(False)
         elif not supports_paged_attn():
             logger.warning(
                 "Flash attention version >=2.5.7 "
@@ -233,6 +234,7 @@ class ExllamaV2Container:
             self.config.no_flash_attn = True
             self.paged = False
             self.max_batch_size = 1
+            torch.backends.cuda.enable_flash_sdp(False)
 
         # Set k/v cache size
         # cache_size is only relevant when paged mode is enabled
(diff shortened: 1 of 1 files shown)


quarterturn:
Same error having made the changes below to backends/exllamav2/model.py:
```
    210                 "To disable compatability mode, all GPUs must be ampere "
    211                 "(30 series) or newer. AMD GPUs are not supported."
    212             )
    213             self.config.no_flash_attn = True
    214             self.paged = False
    215             self.max_batch_size = 1
    216         elif not supports_paged_attn():
    217             logger.warning(
    218                 "Flash attention version >=2.5.7 "
    219                 "is required to use paged attention. "
    220                 "Switching to compatibility mode. \n"
    221                 "This disables parallel batching "
    222                 "and features that rely on it (ex. CFG). \n"
    223                 "Please upgrade your environment by running a start script "
    224                 "(start.bat or start.sh)\n\n"
    225                 "Or you can manually run a requirements update "
    226                 "using the following command:\n\n"
    227                 "For CUDA 12.1:\n"
    228                 "pip install --upgrade .[cu121]\n\n"
    229                 "For CUDA 11.8:\n"
    230                 "pip install --upgrade .[cu118]\n\n"
    231                 "NOTE: Windows users must use CUDA 12.x to use flash-attn."
    232             )
    233             self.config.no_flash_attn = True
    234             self.paged = False
    235             self.max_batch_size = 1
```

turboderp:
@quarterturn Are you using a draft model?

quarterturn:
> @quarterturn Are you using a draft model?

I am not.

quarterturn:
It's working. Thanks for the fix.

PR #158, opened 2024-07-26
feat: add embeddings support via Infinity-emb

AlpinDale:
**Examples**
```sh
curl -X POST http://localhost:5000/v1/embeddings -H "Content-Type: application/json" -d '{ "model": "sentence-transformers/all-mpnet-base-v2", "input": ["I am the text to extract meaning from"] }' | jq .
```

The model needs to be downloaded beforehand. `huggingface-cli download all-mpnet-base-v2` will take care of it (the model's pretty small). All sentence-transformers type of models are supported.

TODO:
- [ ] Optionally import `endpoints.OAI.embeddings`, with a config flag.
- [ ] Write test units.


bdashore3:
Instead of using sentence-transformers, use Infinity-emb which is an async batching engine for embeddings which should prevent any possibility of race conditions.

Inline embedding model loading will be added in a separate PR/set of commits (#130)

PR #175, opened 2024-08-23
Model: Fix draft model loading

veden:
The problem is that when a draft model is added and the cache class is attempting to load the `self.model` object which is `None` and `.config` cannot be found because the model that is loaded at the time of draft cache creation is `self.draft_model`

Who doesn't enjoy extra tokens per second.


veden:
main branch has separate fix, closing as this is no longer needed.

PR #178, opened 2024-08-26
Add healthcheck

SecretiveShell:
Adds:
- localhost only /healthcheck endpoint
- cURL healthcheck in docker compose file

A healthcheck allows the docker daemon to restart tabbyAPI in case of a crash

**Examples**
If the tabbyAPI server crashes and hangs indefinitely then docker will not restart the application without a health check

### DIFF
--- a/endpoints/core/router.py
+++ b/endpoints/core/router.py
@@ -43,13 +43,13 @@ from endpoints.core.utils.model import (
 
 router = APIRouter()
 
+
 # Healthcheck endpoint
-@router.get(
-    "/health",
-)
-async def healthcheck() :
+@router.get("/health")
+async def healthcheck():
     """Get the current service health status"""
     return {"status": "healthy"}
+
 
 # Model list endpoint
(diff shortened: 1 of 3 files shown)


bdashore3:
Couple of things here:
1. For web convention's sake, it's best to make the path `/health` instead of `/healthcheck`
2. This doesn't need to be restricted to localhost. It's a harmless endpoint, so exposing it without auth should be okay.

SecretiveShell:
For point 3 I assume you meant a docstring, if this is incorrect please let me know. Thanks.

endpoints/core/router.py
```diff
@@ -519,3 +519,12 @@ async def unload_sampler_override():
     """Unloads the currently selected override preset"""
 
     sampling.overrides_from_dict({})
+
+
+@router.get(
+    "/health",
+    dependencies=[],
```
bdashore3, line 526: Can remove the `dependencies` part here.

SecretiveShell:
Thanks for the clarification, I have made the changes requested.

PR #184, opened 2024-08-31
Complete conditional infinity import TODO

SecretiveShell:
- additional logging
- changed declaration order prevents has_infinity_emb becoming undefined when there are errors other then import error
- added additional type hinting

**Examples**
The software is more likely to crash without these changes if there is an issue with the infinity embed backend

**Additional context**
completes a TODO in the code

### DIFF
--- a/backends/infinity/model.py
+++ b/backends/infinity/model.py
@@ -7,13 +7,13 @@ from typing import List, Optional
 from common.utils import unwrap
 
 # Conditionally import infinity to sidestep its logger
-# TODO: Make this prettier
+has_infinity_emb: bool = False
 try:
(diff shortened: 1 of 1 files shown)


bdashore3:
I agree with a refactor of the error handling here, but I'd also like to remove the log statements for two reasons:
1. The logger isn't set up at the time of import
2. An import check isn't required unless a user wants to use embeddings, which the codebase already checks for. The import is at the top-level because of infinity's own logger which requires an issue on their end to make it configurable.

I can merge this after removing those logging statements and replacing the exception with a `pass`.

SecretiveShell:
should be ready for merge now, I made all the changes asked.

PR #185, opened 2024-09-04
Refactor config loading

SecretiveShell:
This PR refactors the way config files are loaded into tabbyAPI

It refactors the current functions to return a dictionary, reducing the use of the global variables.
A new load function is added to aggregate all the possible load config functions into a list of dictionaries
The global config is constructed by performing a deep merge on all the resulting configurations in the order they are listed
The individual functions to get the configuration are refactored to use a single factory, with nested functions using nested calls

**Examples**
This makes adding new sources of configuration much easier
The order of which configuration takes precedence when multiple competing settings are present is now much more obvious


bdashore3:
Good concept. However, I'd like a merge function in `common/utils.py` rather than needing to install a new package for a dict merge. I don't believe tabby will get complicated enough to warrant a full kit of merging dictionaries together.

Also, please run `ruff format` which should pass the PR checks. If ruff isn't installed, run `pip install .[dev]` in your venv and try again.

SecretiveShell:
I have added the internal implementation as requested

bdashore3:
This PR wasn't working when testing, but the concept was good. So, I decided to migrate the global config module into a shared class instance instead. This should preserve the flexibility of adding more config options while using class based references in other modules.

PR #194, opened 2024-09-08
add static type checker to dev environment

SecretiveShell:
Currently there is a good chance to make mistakes when contributing new code, for example the invalid arguments in #191, which could have easily been caught before making its way into the production build using static analysis

This feature will work directly to reduce the amount of bugs in the code and drastically help with refactoring code. Here are some example issues that it detected when I was testing the scripts:

endpoints\OAI\utils\embeddings.py:64: error: Incompatible return value type (got "EmbeddingsResponse", expected "dict[Any, Any]")  [return-value]
endpoints\Kobold\router.py:126: error: Incompatible return value type (got "dict[str, Any]", expected "MaxLengthResponse")  [return-value]

**Examples**
see #191 

**Additional context**
moving the config to pydantic as in #189 would expand the functionality in this to prevent issues like #104

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,7 +56,8 @@ extras = [
     "sentence-transformers",
 ]
 dev = [
-    "ruff == 0.3.2"
+    "ruff == 0.3.2",
+    "mypy",
 ]
 cu121 = [
     # Torch (Extra index URLs not support in pyproject.toml)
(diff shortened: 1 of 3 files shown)


kingbri1:
After some testing, Pyright is a better option due to compatibility with editors and ease of setup. Mark this as a goal for later on when it makes sense to look at typing.

However, Python's (unlike TS's) typing is difficult, especially with request params that can be nullish.

Closing.

PR #230, opened 2024-10-30
Remove override_base_seq_len

DocShotgun:
Users frequently modify `override_base_seq_len` improperly despite warnings in the example config not to do so, and problem that `override_base_seq_len` was introduced to solve is a rare edge case.

N/A

**Examples**
N/A

**Additional context**
`override_base_seq_len` was introduced as a workaround to enable automatic rope alpha calculation for models with incorrect/misleading `max_position_embeddings` in the base model config, specifically Mistral-7B-v0.1 with 32768 context specified, however only ~8-9k tokens of effective context when used without SWA. In this situation, you could set `max_seq_len` to 16384, and `override_base_seq_len` to 8192 to enable automatic rope alpha scaling to extend the model to 16384 context.

However this arg proved to just cause confusion for users rather than have any tangible benefit. We now recommend just modifying the model's `max_position_embeddings` in `config.json` as appropriate if you need automatic rope alpha scaling.


bdashore3:
Yeah, it's best to remove this since users don't read the large warning that says "Don't default to using this"

PR #233, opened 2024-11-04
feat: log the current commit hash

AlpinDale:
As per #232, we need a way to figure out which tabby version we're using. Since tabby doesn't have versions, we can use commit hash instead.

This PR creates a commit hash whenever the server is started, and saves it to `endpoints/OAI/_commit.py` (gitignored).


bdashore3:
Rather than embedding this in start.py, I think it would make more sense to place this in a utils function somewhere that's triggered by main. Also, prefer YAML over storing python files.

Issue #234, opened 2024-11-11
[BUG] Tool Calling not working for Llama 3.2 3B


raisbecka:
### OS

Windows

### GPU Library

CUDA 12.x

### Python version

3.11

### Describe the bug

I am using WSL2 docker with CUDA. Regular text generation runs really well and as expected, but tool calling doesn't work even when the exact name of the tool is in the prompt. Using default prompt template that comes with Llama 3.2 exl2 (https://huggingface.co/bartowski/Llama-3.2-3B-Instruct-exl2).

**Here's the thing:** Using Ollama on Windows, if I include a tool list with the prompt, it works correctly 95% of the time. For regular chat, I simply omit the tool list when I send the request - otherwise this smaller model tends to hallucinate. But this workflow works very consistently for me. **This is with the same default prompt template.**

I would prefer to continue using the faster TabbyAPI + ExLlamav2 server, but I need a reliable way to call tools. **I know in advance when a prompt should result in at least one tool call**, so how can I use this to my advantage (like I do with Ollama)?

### Reproduction steps

- Use model: https://huggingface.co/bartowski/Llama-3.2-3B-Instruct-exl2/ (the 3_5 branch).
- Use default prompt template included with model
- Use openai module in Python to connect
- Try regular chat messages with no tools parameter supplied. They work.
- Try tool prompts with the tools param set to a list of properly defined tools. Doesn't work.

### Expected behavior

I should be able to trigger a tool call when I request one and supply a tools param with matching tool names. I should also be able to switch back to normal dialog by supplying an empty tool list.

### Logs

**Below is a log output of the request being sent by openai module:**

[...]

PR #253, opened 2024-11-24
workaround for tool calling

randoentity:
Related to #234.

Not a proper fix, but could be merged into a local branch if anyone needs a quick workaround.

**Examples**
Enables tool calling again. However, it still raises an error when combined with draft models. To work around that background_drop should be called early. I'm having a hard time understanding the code in the time I have, and SD+TC is probably only wanted by a few people, so this is all I have for now.

**Additional context**
I tried disabling the draft model just before generating the prompt for the tool call (mainly to not have to relaunch TabbyAPI to switch from using a draft for coding and back to using tool calling), but I got into a tool call loop. That's not the way to fix this anyway.


bdashore3:
For now, this should be okay. The draft model issue can be tested and fixed at a later date.

PR #254, opened 2024-11-27
add draft_gpu_split option for spec decoding

lucyknada:
with a setup of 3090+3060; loading 32b (5bpw) and 1.5b or 0.5b draft (6bpw) first loads the draft model and then the model; making an otherwise 32b fitting onto the single 3090 (a lot more bandwidth) leak onto the 3060 with quite a few layers, negating all benefits of the draft model and even dropping performance even more (15-30% less tps); now pinning the draft model to another gpu (the 3060) entirely bypasses that and you get the same performance boost for spec decoding as before.

tabby_config.yml example of a draft_gpu_split:

image

I have tested it both with autosplit on the target model and manually splitting via gpu_split, the gpu-split syntax stays the same to be able to re-use the same code.

this is merely a kick-off PR to make it work, as I'm not sure if there's anything else that needs to be modified for this to be upstreamable and if there's some neat python tricks to not have to duplicate code in e.g. initializing gpu_device_list.

### DIFF
--- a/backends/exllamav2/model.py
+++ b/backends/exllamav2/model.py
@@ -89,8 +89,8 @@ class ExllamaV2Container:
     generation_config: Optional[GenerationConfig] = None
 
     # GPU split vars
-    gpu_split: Optional[list] = None
-    draft_gpu_split: Optional[list] = None
+    gpu_split: List[float] = []
+    draft_gpu_split: List[float] = []
     gpu_split_auto: bool = True
     autosplit_reserve: List[float] = [96 * 1024**2]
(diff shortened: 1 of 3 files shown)


lucyknada:
just as I sent this I realized the sample config in the root of the project would need to have this option added too, I'm exclusively an in-line user so it escaped my mind entirely.

kingbri1:
@lucyknada please check the changes and see if they work. They run properly on my end.

lucyknada:
confirming it works! thanks

PR #265, opened 2024-12-21
Adding multimodal tests

christopherthompson81:
I had some difficulty in enabling multimodal functionality, more so than when doing so directly with exllamav2, so I wanted to include their multimodal tests here.

This should help others setting up a multimodal model verify that it is working and provide a demonstration of how to use it.


kingbri1:
These are moreso client-side tests that would make more sense to go in a separate repo. The tests folder hasn't been used or updated for a while and they mainly use tabbyAPI internally rather than connecting as a server.

For vision, would make more sense to add documentation for enabling and using a vision model and should supersede this PR.

Issue #294, opened 2025-03-01
[REQUEST] Add reasoning content parsing


Orion-zhen:
### Problem

I'm playing with DeepSeek-R1-Distill-Llama-70B with Page Assist, when I ask LLM a question, it just generated a bunch of text without splitting it into thinking part and answer part. However, when I use other backends like vllm, it first started to think in a folded thinking box, and then gave the answer outside the box.

### Solution

I have looked through `endpoint/OAI`, but I'm not familiar with `async` programming 🥲. The feature can be implemented by adding a reasoning content parser function, which detects `<think>` or `</think>` tags in LLM's response, and separate them into `reasoning_content` and `content`. And I have to mention that `<think>` and `</think>` tags are not necessarily in perfect pairs. DeepSeek-R1-Distill models have `<think>` tag in chat template, so the model only generate `</think>` tag. In that case, we can just divide content by `</think>` tag.

### Explanation

As DeepSeek released their new R1 models, `reasoning_content` field is added to their API reference. Enable reasoning content parsing is very helpful for chat frontends, because reasoning content can be folded or unfolded to make the whole UI clean and let users focus on the answer. As far as I know, vllm and llama.cpp have already supported this feature.

### Additional context

Thank you for your reading, I would be grateful if you could add this feature to the repo.

### Acknowledgements

- [x] I have looked for similar requests before submitting this one.
- [x] I understand that the developers have lives and my issue will be answered when possible.
- [x] I understand the developers of this program are human, and I will make my requests politely.

strikeoncmputrz:
Bumping this based on MiniMax M2 and M2.1. Neither model includes `<think>` in completions that include reasoning. Would be great to be able to enable / disable `reasoning_content` in TabbyAPI.  I recognize that it doesn't conform to the OpenAI API spec. 

I'm taking the performance hit right now and running Minimax  M2.1 in Llama.cpp.  Started a discussion about this on [r/LocalLlama](https://www.reddit.com/r/LocalLLaMA/comments/1q1gps8/minimax_m21_think_tag_and_interleaved_thinking/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button)

turboderp:
Closing some stale issues (tool calling should be fixed with the new parser.)

PR #295, opened 2025-03-01
:sparkles: reasoning parser support

Orion-zhen:
The PR solves [issue294](https://github.com/theroyallab/tabbyAPI/issues/294)

The PR implemented reasoning parser and streaming reasoning parser, and was tested on DeepSeek-R1-Distill-Llama-70B with frontend like Page Assist and Cherry Studio. In my tests, it worked well.

I added configuration entry points to `config.yml` and placed it under network section.

### DIFF
--- a/common/config_models.py
+++ b/common/config_models.py
@@ -99,6 +99,27 @@ class NetworkConfig(BaseConfigModel):
             "Possible values: OAI, Kobold."
         ),
     )
+    reasoning_parser: bool = Field(
+        False,
+        description=(
+            "Enable the reasoning parser (default: False).\n"
+            "This will split response message into reasoning_content and content fields."
+        )
+    )
+    reasoning_start_token: str = Field(
+        "<think>",
+        description=(
+            "Start token for the reasoning parser (default: <think>).\n"
+            "This token is used to split the response message into reasoning_content and content fields."
+        )
+    )
+    reasoning_end_token: str = Field(
+        "</think>",
+        description=(
+            "End token for the reasoning parser (default: </think>).\n"
+            "This token is used to split the response message into reasoning_content and content fields."
+        )
+    )
 
     # Converts all strings in the api_servers list to lowercase
     # NOTE: Expand if more models need this validator
--- a/config_sample.yml
+++ b/config_sample.yml
@@ -31,6 +31,16 @@ network:
   # Possible values: OAI, Kobold.
   api_servers: ["OAI"]
 
+  # Enable reasoning parser (default: False).
+  # Do NOT enable this if the model is not a reasoning model (e.g. deepseek-r1 series)
+  reasoning_parser: false
+
+  # The start token for reasoning conetnt (default: "<think>")
+  reasoning_start_token: "<think>"
+
+  # The end token for reasoning conetnt (default: "</think>")
+  reasoning_end_token: "</think>"
+
 # Options for logging
 logging:
   # Enable prompt logging (default: False).
--- a/endpoints/OAI/types/chat_completion.py
+++ b/endpoints/OAI/types/chat_completion.py
@@ -31,6 +31,7 @@ class ChatCompletionMessagePart(BaseModel):
 class ChatCompletionMessage(BaseModel):
     role: str = "user"
     content: Optional[Union[str, List[ChatCompletionMessagePart]]] = None
+    reasoning_content: Optional[str] = None
     tool_calls: Optional[List[ToolCall]] = None
     tool_calls_json: SkipJsonSchema[Optional[str]] = None
 
--- a/endpoints/OAI/utils/chat_completion.py
+++ b/endpoints/OAI/utils/chat_completion.py
@@ -17,6 +17,7 @@ from common.networking import (
     handle_request_error,
     request_disconnect_loop,
 )
+from common.tabby_config import config
 from common.utils import unwrap
 from endpoints.OAI.types.chat_completion import (
     ChatCompletionLogprobs,
@@ -33,6 +34,20 @@ from endpoints.OAI.utils.completion import _stream_collector
 from endpoints.OAI.types.tools import ToolCall
 
 
+def _extract_think_content(text: str) -> tuple[Optional[str], Optional[str]]:
+    """Extract content between <think> tags and the remaining content. Only available in none-streaming mode."""
+    if config.network.reasoning_start_token not in text and config.network.reasoning_end_token not in text:
+        return None, text
+    elif config.network.reasoning_start_token in text:
+        start_reasoning = text.split(config.network.reasoning_start_token)[1]
+        reasoning_content = start_reasoning.split(config.network.reasoning_end_token)[0]
+        content = start_reasoning.split(config.network.reasoning_end_token)[1]
+        return reasoning_content, content
+    else:
+        reasoning_content = text.split(config.network.reasoning_end_token)[0]
+        content = text.split(config.network.reasoning_end_token)[1]
+        return reasoning_content, content
+
 def _create_response(
     request_id: str, generations: List[dict], model_name: Optional[str]
 ):
@@ -43,9 +58,16 @@ def _create_response(
 
     choices = []
     for index, generation in enumerate(generations):
-        message = ChatCompletionMessage(
-            role="assistant", content=unwrap(generation.get("text"), "")
-        )
+        if config.network.reasoning_parser:
+            raw_content = unwrap(generation.get("text"), "")
+            reasoning_content, content = _extract_think_content(raw_content)
+            message = ChatCompletionMessage(
+                role="assistant", reasoning_content=reasoning_content, content=content
+            )
+        else:
+            message = ChatCompletionMessage(
+                role="assistant", content=unwrap(generation.get("text"), "")
+            )
 
         tool_calls = generation["tool_calls"]
         if tool_calls:
@@ -103,6 +125,7 @@ def _create_stream_chunk(
     generation: Optional[dict] = None,
     model_name: Optional[str] = None,
     is_usage_chunk: bool = False,
+    is_reasoning_chunk: bool = False,
 ):
     """Create a chat completion stream chunk from the provided text."""
 
@@ -137,6 +160,8 @@ def _create_stream_chunk(
 
     else:
         message = ChatCompletionMessage(
+            role="assistant", reasoning_content=unwrap(generation.get("text"), "")
+        ) if is_reasoning_chunk else ChatCompletionMessage(
             role="assistant", content=unwrap(generation.get("text"), "")
         )
 
@@ -328,6 +353,8 @@ async def stream_generate_chat_completion(
 
         # We need to keep track of the text generated so we can resume the tool calls
         current_generation_text = ""
+        
+        is_reasoning_chunk = config.network.reasoning_parser
 
         # Consumer loop
         while True:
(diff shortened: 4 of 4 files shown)


kingbri1:
Hey @Orion-zhen, thanks for PRing! I see from the issue that Deepseek does extend the OAI API in this way, so it should be ok to add this into Tabby as this looks like a standard.

Can you please send a test request to make sure the PR works? I assume I'd use a deepseek distill to test this out?

Orion-zhen:
Sure! Here are test requests on none-streaming and streaming request:

none-streaming:

```python
from openai import OpenAI
client = OpenAI(api_key="<whatever-key>", base_url="<local-url>")

messages = [{"role": "user", "content": "9.11 and 9.8, which is greater?"}]
response = client.chat.completions.create(
    model="<model-name>",
    messages=messages
)

reasoning_content = response.choices[0].message.reasoning_content
content = response.choices[0].message.content

print(reasoning_content)
print(content)
```

streaming:

```python
from openai import OpenAI
client = OpenAI(api_key="<whatever-key>", base_url="<local-url>")

# Round 1
messages = [{"role": "user", "content": "9.11 and 9.8, which is greater?"}]
response = client.chat.completions.create(
    model="<model-name>",
    messages=messages,
    stream=True
)

reasoning_content = ""
content = ""

for chunk in response:
    if chunk.choices[0].delta.reasoning_content:
        reasoning_content += chunk.choices[0].delta.reasoning_content
    else:
        content += chunk.choices[0].delta.content

print(reasoning_content)
print(content)
```

Orion-zhen:
I added reasoning start token handler, for the latest released model QwQ-32B can explicitly output `<think>` token.

Orion-zhen:
👋 Hi there, any update?

Orion-zhen:
> Any reason these settings are under the network section of the config file? Should they not be model specific?

Hi, @ladrua. Because as far as I think, it's an API level configuration. The standard API can contain `reasoning_content` even if the model is not a reasoning model. For example, you can use system prompt to force the model to think with `<think>` and `</think>` tokens, which should be regarded as `reasoning_content` as well.

ladrua:
Hi :) I see, I just notice that some models use different thinking tokens. some use `thinking`(plural) some use `reasoning`, and since tabby supports model switching, would it not make sense to at least have the thinking tokens defined on a per model basis? Anyway, I have just tested your PR and its working great with qwq-32b. Thank you.

Orion-zhen:
> Hi :) I see, I just notice that some models use different thinking tokens. some use `thinking`(plural) some use `reasoning`, and since tabby supports model switching, would it not make sense to at least have the thinking tokens defined on a per model basis? Anyway, I have just tested your PR and its working great with qwq-32b. Thank you. 

Thank you for testing my PR! I think splitting configuration about the same topic (reasoning) into two different sections would bring inconvenience for users to setup and change configs. As for model switching, tabby provides inline switching and switching with config file. For inline switching, the config file is not changed, which means if the new loaded model uses another reasoning token, you should change it in the config file, too. For config files, I think it's better to put configs about reasoning together.

kingbri1:
Backreading this conversation, I agree with @ladrua here. This pattern has already been established for vision with the `vision: true` flag in config.

The network block is specifically for options pertaining to the server regarding IPs, ports, auth, etc. Reasoning doesn't seem to be a good fit here as that depends on the model. If users really want reasoning support from a non-reasoning model, then add it in the model block or inline config.

Orion-zhen:
Sure, I will change it then.

Orion-zhen:
Sure, I will change it then.

Orion-zhen:
@ladrua and @kingbri1, thank you for your insights.

Originalimoc:
@Orion-zhen
Can this now be configured in tabby_config.yml? That is where per model config resides.
"reasoning" set in global but "reasoning_start_token" "reasoning_end_token" should be in model config.
And there should be a detect of chat template if the <reasoning_start_token> is really there to prevent undesired adding/parsing.

Orion-zhen:
> @Orion-zhen Can this now be configured in tabby_config.yml? That is where per model config resides. "reasoning" set in global but "reasoning_start_token" "reasoning_end_token" should be in model config, heck, put all three into per model config. And there should be a detection of chat template if the <reasoning_start_token> is really there to prevent undesired adding/parsing. Add the token(add none if none) to the full response then parsing is a better idea.

Hi @Originalimoc.

1. this config can now be configured in `config.yml` in the `model` section, just like `vision` flag.
2. `reasoning`, `reasoning_start_token` and `reasoning_end_token` are all placed in per model config
3. I'm currently not considering to check whether `reasoning_start_token` is in the chat template or not. I have warned users in `config_sample.yml` not to enable `reasoning` if the model is not a reasoning model. Checking chat template might further increase code complexity. Similar implementations like vllm won't check chat template, too.

Orion-zhen:
Although current reasoning models such as R1, QwQ-32B, etc. have `reasoning_start_token` in their chat template, it's uncertain if future models can naturally think without explicitly putting `reasoning_start_token` in chat template. Detection of `reasoning_start_token` in chat template might not be a good criterion.

Originalimoc:
1/ I mean in model_folder(like Qwen2-VL-7B)/tabby_config.yml, which look like this:
image

2/ That is why this should be put into per model config "tabby_config.yml" not global, you can not check chat template this way just manually set it in this config file.
3/ QwQ usually starts outputting its <think> after <assistant>. Some sizes of R1 distilled do that too. But it might just output the answer without <think></think>, chat template is just a steering method.

Orion-zhen:
I'm not familiar with this per model config feature. TBH, this is the first time that I hear of it. I think reasoning feature is something equal to vision feature, putting it along with `vision` flag should make sense. Just as your comments, `vision` flag should be per model config too, but it still appears in global `config.yml`. According to your comment, per model config is kind of buggy, so I would better keep this config in `config.yml` before the feature becomes stable. And I would focus this PR on reasoning parser feature.

Originalimoc:
It's just missing some field when merging, after merging it's processed by the same config code. I actually almost configured nothing in global config.yml because I use inline loading all the time and each model requires entirely different ctx length/cache mode/draft model/vision etc.

ladrua:
@Orion-zhen Maybe we should check for `\n` or whitespace in the beginning of the splits? When I use this with QwQ 32b it always starts with a empty new line, both for the reasoning tokens and the normal content tokens.

Orion-zhen:
> @Orion-zhen Maybe we should check for `\n` or whitespace in the beginning of the splits? When I use this with QwQ 32b it always starts with a empty new line, both for the reasoning tokens and the normal content tokens.

In none-streaming cases, we can simply `strip` contents. But in streaming cases, checking for whitespace can be tricky. The model generates token per token, so we should track reasoning start/end token as well as next token(s). The whole streaming logic is implemented in a loop, so I would have to add more variables outside the loop to record status. I'm considering if it's worthy to add more complexity for this feature. Because in my use cases, chatbot apps like cherry-studio and chatbox can display messages normally even though there are whitespaces. Do you have any good idea? 🤔

Originalimoc:
@ladrua Try add \n to the chat template after "\<think\>"?/enable token healing?

Originalimoc:
After this PR get merged I'll create a new PR fixing the per model config loading.

ladrua:
> Do you have any good idea? 🤔

Something like this would work for the streaming bit:
```python
async def stream_generate_chat_completion(
    prompt: str,
    embeddings: MultimodalEmbeddingWrapper,
    data: ChatCompletionRequest,
    request: Request,
    model_path: pathlib.Path,
):
    """Generator for the generation process."""
    abort_event = asyncio.Event()
    gen_queue = asyncio.Queue()
    gen_tasks: List[asyncio.Task] = []
    disconnect_task = asyncio.create_task(request_disconnect_loop(request))

    try:
        logger.info(f"Received chat completion streaming request {request.state.id}")

        for n in range(0, data.n):
            task_gen_params = data.model_copy(deep=True)

            gen_task = asyncio.create_task(
                _stream_collector(
                    n,
                    gen_queue,
                    prompt,
                    request.state.id,
                    abort_event,
                    embeddings=embeddings,
                    **task_gen_params.model_dump(exclude={"prompt"}),
                )
            )

            gen_tasks.append(gen_task)

        # We need to keep track of the text generated so we can resume the tool calls
        current_generation_text = ""

        is_reasoning_chunk = config.model.reasoning
        # Add a flag to track if this is the first token
        is_first_token = True

        # Consumer loop
        while True:
            if disconnect_task.done():
                abort_event.set()
                handle_request_disconnect(
                    f"Chat completion generation {request.state.id} cancelled by user."
                )

            generation = await gen_queue.get()
            # lets only append the text if we need it for tool calls later
            if data.tool_call_start and "text" in generation:
                current_generation_text += generation["text"]

            # check if we are running a tool model, and that we are at stop
            if data.tool_call_start and "stop_str" in generation:
                generations = await generate_tool_calls(
                    data,
                    [generation],
                    request,
                    current_generations=current_generation_text,
                )
                generation = generations[0]  # We only have one generation in this case

            # Stream collector will push an exception to the queue if it fails
            if isinstance(generation, Exception):
                raise generation

            if (
                unwrap(generation.get("text"), "") == config.model.reasoning_start_token
                and config.model.reasoning
            ):
                # Update reasoning chunk flag
                is_reasoning_chunk = True
                # And skip this token
                continue
            if (
                unwrap(generation.get("text"), "") == config.model.reasoning_end_token
                and config.model.reasoning
            ):
                # Update reasoning chunk flag
                is_reasoning_chunk = False
                # And skip this token
                continue

            # Strip whitespace from the first token if it contains text
            if is_first_token and "text" in generation and not is_reasoning_chunk:
                # Apply lstrip() to the first token
                generation["text"] = generation["text"].lstrip()
                logger.info(f"First token stripped for request {request.state.id}")

            response = _create_stream_chunk(
                request.state.id,
                generation,
                model_path.name,
                is_reasoning_chunk=is_reasoning_chunk,
            )
            yield response.model_dump_json()

            # After processing the first token, set the flag to False
            if is_first_token and not is_reasoning_chunk:
                is_first_token = False

            # Check if all tasks are completed
            if all(task.done() for task in gen_tasks) and gen_queue.empty():
                # Send a usage chunk
                if data.stream_options and data.stream_options.include_usage:
                    usage_chunk = _create_stream_chunk(
                        request.state.id,
                        generation,
                        model_path.name,
                        is_usage_chunk=True,
                    )
                    yield usage_chunk.model_dump_json()

                logger.info(
                    f"Finished chat completion streaming request {request.state.id}"
                )

                yield "[DONE]"
                break
    except CancelledError:
        # Get out if the request gets disconnected

        if not disconnect_task.done():
            abort_event.set()
            handle_request_disconnect("Chat completion generation cancelled by user.")
    except Exception:
        yield get_generator_error(
            "Chat completion aborted. Please check the server console."
        )
```

mratsim:
> So I've been testing the hell of this since I quantized GLM-4.7.
> 
> I have a suspicion that if `<think>` or `</think>` are generated within the text by the LLM, they are always replaced. But if you work on a rewrapper that converts back `reasoning_content` into `<think>reasoning_content</think>` (to solve [Vali-98/ChatterUI#384](https://github.com/Vali-98/ChatterUI/issues/384)), the textual `<think>` tokens seems problematic:
> image tabbyapi_think_token
> 
> I'm not sure how vllm/sglang solve this, it seems difficult to work directly at the token level, say: https://huggingface.co/mratsim/GLM-4.7-EXL3/blob/3.84bpw-tuned/tokenizer_config.json#L171-L186
> 
> ```
>     "151350": {
>       "content": "<think>",
>       "lstrip": false,
>       "normalized": false,
>       "rstrip": false,
>       "single_word": false,
>       "special": false
>     },
>     "151351": {
>       "content": "</think>",
>       "lstrip": false,
>       "normalized": false,
>       "rstrip": false,
>       "single_word": false,
>       "special": false
>     },
> ```
> 
> And this might get even more tricky with GLM-4.7 Preserved Thinking and Interleaved Thinking.
> 
> But even just supporting a single thinking/reasoning content section for starters and not treating second and third reasoning can solve 99% of the use cases.
> 
> Lastly, we probably need a configuration to output in either `reasoning_content` (historical DeepSeek R1 API) or `reasoning` (new OpenAI and OpenRouter recommendation: [vllm-project/vllm#27755](https://github.com/vllm-project/vllm/issues/27755))

So there is no fix in vLLM / SGLang :shrug:, and it's such a niche use-case, at least until interleaved thinking gets generalized so I think it's fine.

Issue #304, opened 2025-03-14
[BUG] Model Switching Fails with Concurrent Requests


bold84:
### OS

Linux

### GPU Library

CUDA 12.x

### Python version

3.12

### Describe the bug

When attempting to switch models implicitly through API requests, TabbyAPI fails with multiple errors:
1. `rich.errors.LiveError: Only one live display may be active at once` - This occurs when multiple model loading operations are attempted simultaneously
2. `torch.OutOfMemoryError: CUDA out of memory` - The system runs out of GPU memory during model loading
3. `AttributeError: 'NoneType' object has no attribute 'bos_token'` - The tokenizer is not properly initialized when trying to use a model

This does not always happen. 

### Reproduction steps

1. Start TabbyAPI with a model loaded
2. Send a chat completion request specifying a different model than the currently loaded one
3. Send additional requests in quick succession that also specify different models
4. Observe the errors in the logs

### Expected behavior

TabbyAPI should:
1. Properly unload the current model before loading a new one
2. Handle concurrent model loading requests by either queuing them or rejecting additional requests with a clear error while a model is being loaded
3. Ensure tokenizers are properly initialized before attempting to use them
4. Provide clear error messages to the client when model switching fails

### Logs

```
tabbyapi-1  | 2025-03-14 11:05:06.977 INFO:     Unloading existing model.
tabbyapi-1  | 2025-03-14 11:05:08.884 INFO:     Model unloaded.
tabbyapi-1  | 2025-03-14 11:05:08.898 INFO:     Attempting to load a prompt template if
tabbyapi-1  | present.
tabbyapi-1  | 2025-03-14 11:05:08.900 INFO:     Attempting to load a prompt template if
tabbyapi-1  | present.
tabbyapi-1  | 2025-03-14 11:05:08.907 INFO:     Using template "qwq" for chat completions.
tabbyapi-1  | 2025-03-14 11:05:09.064 INFO:     Using template "qwq" for chat completions.
tabbyapi-1  | 2025-03-14 11:05:09.064 INFO:     172.19.0.1:33056 - "POST /v1/chat/completions
tabbyapi-1  | HTTP/1.1" 500
tabbyapi-1  | 2025-03-14 11:05:09.125 INFO:     Loading draft model:
tabbyapi-1  | /app/models/Qwen2.5-0.5B-Instruct_exl2_6.0bpw
[...]

PR #305, opened 2025-03-14
Fix Concurrent Model Switching Issues and Improve API Request Handling

bold84:
This PR fixes the model switching issues described in [Issue #304](https://github.com/theroyallab/tabbyAPI/issues/304). The current implementation has several problems that cause errors when attempting to switch models through API requests:

1. Rich progress bar conflicts causing "Only one live display may be active at once" errors
2. Incomplete detection of when generation jobs are finished
3. Lack of proper synchronization between concurrent model load requests
4. Critical race condition where tokenizer is unloaded while still being referenced by generation tasks (causing 'NoneType' object has no attribute 'bos_token_id' errors)
5. Missing clear error messages when model switching isn't possible

Reliable model switching is essential for production deployments of TabbyAPI where different clients might need different models. The current implementation causes errors, CUDA out-of-memory problems.

This fix adds:
1. A way to skip progress bars for API-initiated model loads
2. Proper detection of active generation jobs
3. Configurable behavior to either wait for generations to complete or reject immediately
4. Improved error handling with informative messages including model names

**Examples**
Before this fix, if two clients try to switch models concurrently:
```
tabbyapi | rich.errors.LiveError: Only one live display may be active at once
tabbyapi | AttributeError: 'NoneType' object has no attribute 'bos_token'
tabbyapi | torch.OutOfMemoryError: CUDA out of memory
```

After this fix:
1. When `queue_model_switch_requests: false` (default):
```
tabbyapi | ERROR: Sent to request: Cannot switch models while active generations are in progress with model 'ModelA'. Requested model 'ModelB' will not be loaded. Please try again later.
```

2. When `queue_model_switch_requests: true` and all generations complete:
```
tabbyapi | INFO: Request to switch from 'ModelA' to 'ModelB' is waiting for 3 active generations to complete.
tabbyapi | INFO: All active generations completed. Proceeding with model switch to 'ModelB'.
tabbyapi | INFO: Successfully switched to model 'ModelB'
```

**Additional context**
This PR adds a new configuration option `queue_model_switch_requests` (default: false) that controls whether model switch requests should wait for active generations to complete or be rejected immediately.

Implementation details:
- Uses safer attribute checks to prevent NoneType errors
- Adds timeout protection to prevent indefinite waiting
- Uses proper locking to prevent race conditions
- Disables Rich progress displays for API-initiated model loads
- Adds improved error reporting and logging

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,10 @@
 > 
 >  Need help? Join the [Discord Server](https://discord.gg/sYQxnuD7Fj) and get the `Tabby` role. Please be nice when asking questions.
 
+> [!NOTE]
+> 
+> Want to run GGUF models? Take a look at [YALS](https://github.com/theroyallab/YALS), TabbyAPI's sister project.
+
 A FastAPI based application that allows for generating text using an LLM (large language model) using the [Exllamav2 backend](https://github.com/turboderp/exllamav2)
 
 TabbyAPI is also the official API backend server for ExllamaV2.
--- a/docs/01.-Getting-Started.md
+++ b/docs/01.-Getting-Started.md
@@ -6,7 +6,7 @@ To get started, make sure you have the following installed on your system:
   
     - Do NOT install python from the Microsoft store! This will cause issues with pip.
     - Alternatively, you can use miniconda or uv if it's present on your system.
-  
+
 
 > [!NOTE]
 > Prefer a video guide? Watch the step-by-step tutorial on [YouTube](https://www.youtube.com/watch?v=03jYz0ijbUU)
@@ -40,16 +40,16 @@ To get started, make sure you have the following installed on your system:
 
 ### For Advanced Users
 
-5. Follow steps 1-2 in the [For Beginners](#for-beginners) section
-6. Create a python environment through venv:
+1. Follow steps 1-2 in the [For Beginners](#for-beginners) section
+2. Create a python environment through venv:
     1. `python -m venv venv`
     2. Activate the venv
         1. On Windows: `.\venv\Scripts\activate`
         2. On Linux: `source venv/bin/activate`
-7. Install the pyproject features based on your system:
+3. Install the pyproject features based on your system:
     1. Cuda 12.x: `pip install -U .[cu121]`
     2. ROCm 5.6: `pip install -U .[amd]`
-8. Start the API by either
+4. Start the API by either
     1. Run `start.bat/sh`. The script will check if you're in a conda environment and skip venv checks.
     2. Run `python main.py` to start the API. This won't automatically upgrade your dependencies.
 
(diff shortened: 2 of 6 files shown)


bold84:
if `queue_model_switch_requests` is false, the user will be informed why a model switch is currently not possible. 

CleanShot 2025-03-14 at 20 29 24@2x

bold84:
> Huh, okay. So max_seq_length and cache_size are set in the tabby_config.yml within the models exl2 folder, but its ignoring them and setting the model to the max size possible.
> 
> 
> 
> That's... irritating.

Try commenting out these values in the main config.yaml

It might be that the values are not overridden anymore, which would be a bug.

bold84:
> Bug #2, the regular API key stops working because it cant switch the model with it, even if the model is the exact same model already loaded.

Please elaborate...

bold84:
> It works (kinda) but I was for some reason running into out of memory issues on 2x3090 when running 21k Q8 context / 42k cache on 4.0 bpw 70B models.
> 
> 
> 
> Which okay, fair, but they fit just fine when launching with start.sh.
> 
> 
> 
> Does it take more vram to switch models?

Exllama doesn't appear to release all VRAM (keeps a few 100 MB after unloading), so if you went to the maximum available VRAM with your configuration, try reducing the context size slightly.

I initially did solve this problem by aggressively using torch methods, but that was not a good solution as tabbyapi shouldn't know about torch.

bold84:
Great. I am still not sure what exactly you mean with "When you use the regular API key. 
Can you explain how I can reproduce the issue?

bold84:
> > Great. I am still not sure what exactly you mean with "When you use the regular API key. Can you explain step by step how I can reproduce the issue?
> 
> In open-webui when you add tabbyapi as a OpenAI backend use the regular API key instead of the admin key.
> 
> The problem happens every time.
> 
> The loaded model should work even with a regular API key, but it does not. It lists the error in my screenshot instead.
> 
> This makes having a regular API key 100% pointless, since its not usable.
> 
> Default config with the following changes:
> 
> host: 0.0.0.0 disable_fetch_requests: true model_dir: /mnt/nvme1n1p1/exl2 inline_model_loading: true model_name: Legion-V2.1-LLaMa-70B-4.0bpw-h8-exl2 max_seq_len: 21504 tensor_parallel: true cache_mode: Q8 cache_size: 43008

Thank you. Now I understand.

bold84:
Hi @frenzybiscuit 

I was located in Bangkok and we had an earthquake here today... 
I will get to look into this in the next few days.

bold84:
@frenzybiscuit unfortunately not.
It is quite a big change and as far as I remember, the maintainer wasn't too happy about it and was of the opinion that the goal could be achieved without such heavy changes.

PR #308, opened 2025-03-15
Support more common tool variables in templates (tools, message.tool_calls)

theasp:
Increase the compatibility with VLLM templates which use a non-JSON tools object.

VLLM templates have tools in `tools` as an object, instead of `tools_json` as a JSON string.  I'm trying to get tool calls to work with Mistral, this gives me a known-good template.

Providing the same variable as VLLM would allow using VLLM templates with little or no modification.

**Examples**
- The Mistral parallel tools template from VLLM works up to generating the available tools, provided `if tool_calls is not none` wraps the `[TOOL CALLS]` section to prevent an exception.

Template: https://gist.github.com/theasp/f6574b7e39f92cb210b6204a973fd71a

**Additional context**
- This doesn't fix tool calls in Mistral, the model will generate the tool call, but TabbyAPI doesn't seem to recognize it.


kingbri1:
@theasp thanks for making this PR!

Has VLLM's become the standard for tool calling templates? (Like HF with chat templates)

If so, I'd like to switch tabby's template parsing to adopt to VLLM which ensures compatibility.

theasp:
> Has VLLM's become the standard for tool calling templates? (Like HF with chat templates)
> 
> If so, I'd like to switch tabby's template parsing to adopt to VLLM which ensures compatibility.

I don't think it's become a standard, but the only differences I see between tabby and vllm is that it has the `tools` variable (as in this PR) and that `message.tool_calls` isn't defined if there wasn't a tool call.

Tabby also passes the tool start/stop token as variables in the template, where as vllm has different tool call parsers for different types of models. I'm not sure how Ollama/llama.cpp detect the tool calls.

kingbri1:
I know both ollama and llamacpp do it in weird ways where they have their own templating system that isn't standard, so I'll disregard those. Good to know about VLLM though. Will have to test this because the reason tools_json exists is because its a redirect for the tools param since the JSON needs to be held. Please make sure to test this and I'll do the same on my end.

theasp:
I've discovered that QwQ's default template also uses `tools`:  https://huggingface.co/Qwen/QwQ-32B/blob/main/tokenizer_config.json#L230

```jinja
    {%- for tool in tools %}
        {{- "\n" }}
        {{- tool | tojson }}
    {%- endfor %}
```

I added this to the start of the template:
```jinja
{%- set tool_start = "<tool_call>" %}
{%- set tool_end = "</tool_call>" %}
```

And it looks like it was about to generate a tool call:

image

theasp:
I will be pushing the TabbyAPI changes for this soon, but I have the official qwq template (`qwq-orig-updated.jinja`, they updated the config file on HF) working with only the following changes:

```diff
--- qwq-orig-updated.jinja      2025-03-17 13:38:19.281272639 -0300
+++ qwq-updated.jinja   2025-03-18 00:15:24.421002952 -0300
@@ -1,3 +1,7 @@
+{#- Metadata #}
+{%- set tool_start = "<tool_call>" %}
+{%- set tool_end = "</tool_call>" %}
 {%- if tools %}
     {{- '<|im_start|>system\n' }}
     {%- if messages[0]['role'] == 'system' %}
@@ -59,5 +63,6 @@
 {%- endfor %}
 {%- if add_generation_prompt %}
-    {{- '<|im_start|>assistant\n<think>\n' }}
+    {#- {{- '<|im_start|>assistant\n<think>\n' }} #}
+    {{- '<|im_start|>assistant\n' }}
 {%- endif %}
-
```

Some things I've noticed:
- I can't get the TabbyAPI ChatML tools template to work with Llama 3.1 8b, with or without modifications.  This may be my own fault somehow.
- I can only get QwQ's tool calling to work when using LiteLLM proxy in between OpenWeb-UI and TabbyAPI.  I'm unsure why this makes a difference.
- The addition of `<think>` in the generation prompt from the default table causes the tool calls to not work.  This `<think>` in the generation prompt also causes OpenWeb-UI to not detect that it's thinking as well because the tag is not returned.
- Using the switch template endpoint appears to not clear the changes to `stop_conditions`, containing previous and duplicates `['<tool_call>', '<tool_call>', '<tool_call>', 151645, 151643]`.

@kingbri1  Could you explain your concern about `tools_json` persisting?

kingbri1:
Sure @theasp. This is just theoretical and probably not a concern when I look into it, but in the utils code, `apply_chat_template` is called multiple times since tool calls require > 1 generation when a tool start token is fired. So, there could be confusion of what the tools var is.

Your PR should be fine, but it's just a thought of concern and possible future deprecation.

theasp:
I can get my Mistral template to work under the following conditions:
- LiteLLM is used as a proxy
- `decode_special_tokens` is forced to `True` here: 
https://github.com/theroyallab/tabbyAPI/blob/ccf23243c13769409603a531306cdcbdd7fcac09/backends/exllamav2/model.py#L1386
- I tell the model to say something before the tool call, i.e. "What time is it?  Tell me before you use a tool."

This detects the tool call every time, and I think it even worked once or twice. Usually the model generates forever when processing the tool result.

As an alternative, I forced `stop_conditions` to include `9` (`[TOOL_CALLS]` in Mistral) which didn't help, so I don't think it's a name to number thing.

I suspect that something like the following is going on:
- Special tokens don't function properly as stop tokens when `decode_special_tokens` is `False`.
- The first generated token is discarded if it is a stop token or maybe any special token.

This is getting a bit deeper than just the template variables.

@kingbri1 are you ok with 442345de3737e70fa2fd916c722549420e6bc442 or should I rework `ToolCallProcessor`?

kingbri1:
Fixed some methods that weren't working properly. I also tried the mistral tool call template and it did not work well. I believe this is just due to the model not picking up on it because when I used the chatml template, the tool calls worked fine.

Also, please remove the chat template from this PR. That goes in [llm-prompt-templates](https://github.com/theroyallab/llm-prompt-templates) instead.

I'm ok with what's in that previous commit. I assume this is necessary because the returned `tool_calls` needs to be a dict for the `tools` variable in templates since it functions a bit differently from `tools_json`. If you want to rework ToolCallProcessor, go ahead.

theasp:
I'd like to find a model that's official template uses `message.tool_call_id` (as Mistral does) and works somewhat reliably to figure out #310 as it seems related to this. Do you have any quick ideas about the `[TOOL_CALLS]` token not being detected?  I'll open a bug for it otherwise. 

> Fixed some methods that weren't working properly. I also tried the mistral tool call template and it did not work well. I believe this is just due to the model not picking up on it because when I used the chatml template, the tool calls worked fine.

What model and client specifically are you using with the chatml template that works? I'd love to reproduce it working.

> Also, please remove the chat template from this PR. That goes in [llm-prompt-templates](https://github.com/theroyallab/llm-prompt-templates) instead.

Sure thing, I'll hold off on the PR there until this is merged then.

> I'm ok with what's in that previous commit. I assume this is necessary because the returned `tool_calls` needs to be a dict for the `tools` variable in templates since it functions a bit differently from `tools_json`. If you want to rework ToolCallProcessor, go ahead.

Exactly.  I've added `ToolCallProcessor.dump()` to avoid parsing the just generated JSON.

kingbri1:
I don't believe the tool call token is a bug. This seems to be a model thing since when I prompted with the chatml template, tools were called almost every time. The model I used was the same as you, mistral small 2401.

The 2401 model is pretty bad anyways. I wouldn't be surprised if something is broken there when attempting tool calls/prompting. Let's hold off on making that a bug and I'm happy to discuss on Discord.

As for tool_call_id, is it really needed? The ID is usually randomly generated by the model and from what I can see, it doesn't have much significance.

Please note that this PR is to adopt VLLM's established "standard". I'm strongly against adding special cases and bloating the code because one template has one variable that isn't present. `template_variables` exists for this reason.

theasp:
Ok then, let me know if there is anything further you want done for this PR.

anunknowperson:
изображение
Hello, tried this PR and provided template with Mistral 2501, but it seems to not work at all. I can't catch function call no matter what I do. Any ideas? Probably it should work even if model doesn't support them.

Btw, there is no mention of tool calling in Mistral 2501 huggingface page/official template/docs, but there were mentions of it in previous model. And in previous mistral small there is template withfunction calling. With newest mistral 3.1 there are mentions of function calling, but no official prompt template for it which is kinda strange.

Anyways, please recommend me best way to do function calling in tabby with mistral 2501 or 3.1

Maybe i'm doing something wrong?

anunknowperson:
Tried with chatml, doesn't seem to work even (and thats should be expected, probably?)

anunknowperson:
Maybe I should switch to Qwen 32b with chatml? But its slower and older, I wanted to stick with newer Mistrals.

@kingbri1 sorry for tag, but maybe any advices how to get tool calling to work?

anunknowperson:
изображение
Tried qwen 32b with built-in chatml function calling template and it still doesn't work :(

anunknowperson:
изображение
Interstingly, it does work with litellm proxy, as OP suggested. Any way to fix tools without litellm that for some reason works?

theasp:
@anunknowperson FYI the only model I got working was QwQ. For Mistral, I'm fairly confident that the template portion is correct, but the tool start token is being ignored. 

I've not looked into what litellm is doing to the requests to make them work.

anunknowperson:
Mistrals doesn't work even with litellm proxy. Probably will try to use completion without chat endpoint and parse tools by myself.

davidallada:
I am currently working on overhauling the TabbyAPI tool calling system. I should have a draft up today or next weekend

kingbri1:
Ok this PR is getting a bit out of hand. I'd like to clarify a few things:

1. The original scope of this PR was to add VLLM variables. Which I believe were added and will be merged soon.
2. Tool calling at large isn't standardized (unlike chat templates themselves), especially on the template size. The project keeps itself less bloated by not supporting special cases.
3. I am not trying to support every model that comes with a new template. It will again bloat the code.

I understand that more models come with tool calls, but I am not keen on supporting every possible combination of template. Please adhere to a standard or use the one present here to PR templates to the LLM template repo.

If anyone would like to make PRs with this in mind (such as the tool calling overhaul), I'd be happy to look at them.

PR #320, opened 2025-04-20
Fix RoPE Ratio

Vhallo:
Fixes the RoPE ratio.


kingbri1:
Wanted someone to check this. Ty!

PR #327, opened 2025-05-01
Exl3: Sampler builder

randoentity:
Sampler builder was missing. Probably double work.

Separation of concerns when applying the stack to the sampler (instead of applying it in model.py).

**Examples**
Doesn't work without this :)

**Additional context**
No hard feelings if you pick your own implementation 😄


randoentity:
Still have to add the penalty range I see.

randoentity:
I had to comment out the FA checks as they don't seem implemented yet.
https://github.com/randoentity/tabbyAPI/tree/exl3-hardware-workaround for those wanting a quick patch.
I'm not planning on implementing that right now.

Next thing I can't wait for is TP, but that goes way over my head.

kingbri1:
This was because some untracked files weren't committed. Closing this PR as they are now in the branch.

PR #330, opened 2025-05-02
Model: fix add_bos_token

randoentity:
TypeError('unwrap() takes from 1 to 2 positional arguments but 3 were given')

Just a small fix.

**Examples**
It works.

**Additional context**
There are already some more defaults (True) for this setting, so maybe falling back on another default is superfluous.


kingbri1:
XD thanks

PR #331, opened 2025-05-02
Sampler preset change when inline loading model

xev-c:
**Description**
A bit messy addition to `inline_loading`, with sampler override preset specified in `tabby_config.yml`.
Generation params passed to endpoint in request are kept unchanged, but values not specified are updated for newly loaded model.

This PR allows setting sampler override preset inside models `tabby_config.yml`
When using various models for different tasks as agents, the samplers will switch dynamically too.

**Examples**
Make new sampler override preset inside `sampler_overrides` for example `gemma-3.yml`
Add to `tabby_config.yml` in gemma-3 model directory following line:
```yml
model_sampler_preset: gemma-3
```
Now when gemma-3 is loaded inline samplers will be updated to match what was set for this model.


kingbri1:
It seems like this PR includes multiple features and rewrites a lot of the codebase, which isn't part of the description. Please point from a different branch and only include the proposed changes.

PR #340, opened 2025-05-08
API: accept enable_thinking template variable (to disable thinking with Qwen3)

randoentity:
When a chat completion request defines the enable_thinking variable it will be applied to the chat template.

Adds support for the qwen3 enable_thinking template variable.

Allows to more cleanly disable thinking. As opposed to using /no_think, which still results in an empty <think> tag, this method injects the think tag in the response, which prevents generation of the reasoning block.

**Examples**
Tested on the exl3 branch with this branch merged on top.

**Additional context**
Preserves original thinking when the variable is not sent or the template doesn't check for the variable.

### DIFF
--- a/endpoints/OAI/types/chat_completion.py
+++ b/endpoints/OAI/types/chat_completion.py
@@ -65,6 +65,7 @@ class ChatCompletionRequest(CommonCompletionRequest):
     messages: List[ChatCompletionMessage] = Field(default_factory=list)
     prompt_template: Optional[str] = None
     add_generation_prompt: Optional[bool] = True
+    enable_thinking: Optional[bool] = None
     template_vars: Optional[dict] = {}
     response_prefix: Optional[str] = None
     model: Optional[str] = None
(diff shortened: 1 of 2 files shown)


kingbri1:
For this case, I'd recommend using `template_vars` instead. I've recently added `chat_template_kwargs` as an alias to `template_vars`, so it can be prompted like [VLLM](https://qwen.readthedocs.io/en/latest/deployment/vllm.html#thinking-non-thinking-modes) as stated in Qwen's documentation.

Here's a sample request:
```json
{
    "messages": [
        {
            "role": "user",
            "content": "Hi, how are you?"
        }
    ],
    "max_tokens": 150,
    "template_vars": {
        "enable_thinking": false
    }
}
```

Closing this PR because TabbyAPI does not handle model-specific variables or configurations and adheres to existing standards to reduce bloat. Custom template variables are also documented in the [Wiki](https://github.com/theroyallab/tabbyAPI/wiki/04.-Chat-Completions#template-variables).

If this does become a standard, then this PR can be reopened and the parameter can be added.

PR #364, opened 2025-07-06
Streamline tool calling

kingbri1:
Trims the extra variables from the tool calling implementation while still using template metadata as the primary source of truth for tokens (ex. tool_start).

Now, tool calling in TabbyAPI should be compatible with many more templates that follow the standard of using a tool_start token.

### DIFF
--- a/common/templating.py
+++ b/common/templating.py
@@ -4,6 +4,8 @@ import traceback
 import aiofiles
 import json
 import pathlib
+from dataclasses import dataclass, field
+from datetime import datetime
 from importlib.metadata import version as package_version
 from typing import List, Optional
 from jinja2 import Template, TemplateError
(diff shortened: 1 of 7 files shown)


QuentinDrt:
Hi @kingbri1,

Thanks for this refacto and all the work you've done.

I couldn't use tool calling with Qwen3, one of the reasons is that it requires a different tool_end (</tool_call>.
So as you were working on tool_calling, I sent a PR with the following improvements on your tool-calls branch:

- Adds an optional tool_end variable to the chat_template metadata to handle tool_end tokens like </tool_call>
- Improves ToolCallProcessor by automatically generating type and id so the model can focus on name and function attributes.
- Makes ToolCallProcessor a bit more robust, handling cases where the model returns an object with a function attribute and cases where it directly returns name and arguments.
- Improves generate_tool_calls to better handle parsing of tool_calls and also handle parallel tool_calls.
- Round usage metrics returned in chat_completion response.

Here is the link of the PR: https://github.com/theroyallab/tabbyAPI/pull/365
Let me know if you have any questions about it.

PR #372, opened 2025-08-02
Model: Basic Exl3 TP

randoentity:
Nope. Unless lack of TP in exl3 was a problem.

TP!

**Examples**
Without it tps was 19, with it tps was 24.

**Additional context**
This works, but isn't a serious PR. More work needs to be done.

There's still an error when shutting down. I tried to mitigate an infinite (?) loop by adding the lifespan.
There's also output from the backend that TabbyAPI should probably handle instead. And there's probably some new options to expose.

I'm not sure if the other two tiny changes I did make sense and if they break anything else.

- [ ] fix progress callback
- [ ] remove `progressbar=True`
- [ ] fix unloading

### DIFF
--- a/endpoints/server.py
+++ b/endpoints/server.py
@@ -19,8 +19,11 @@ from endpoints.core.router import router as CoreRouter
 @asynccontextmanager
 async def lifespan(app: FastAPI):
     yield
-    torch.cuda.synchronize()
-    dist.destroy_process_group()
+    try:
+        torch.cuda.synchronize()
+        dist.destroy_process_group()
+    except AssertionError:
+        pass
 
 
 def setup_app(host: Optional[str] = None, port: Optional[int] = None):


backends/exllamav3/model.py
```diff
@@ -450,9 +455,11 @@ def load_model_sync(self, progress_callback=None):
                     yield value
 
         for value in self.model.load_gen(
+            tensor_p=self.use_tp,
             reserve_per_device=self.autosplit_reserve,
             use_per_device=self.gpu_split,
             callback=progress_callback,
+            progressbar=True,
```
kingbri1, line 462: iirc, this arg is for exl's progressbar. Since tabbyAPI uses its own, best to remove this.

turboderp:
Also I fixed the progress bar issue. It was a dumb copy/paste error on my part

endpoints/server.py
```diff
@@ -13,6 +16,16 @@
 from endpoints.core.router import router as CoreRouter
 
 
+@asynccontextmanager
+async def lifespan(app: FastAPI):
+    yield
+    try:
+        torch.cuda.synchronize()
+        dist.destroy_process_group()
```
turboderp, line 24: Is this needed? Calling unload() on the model is supposed to destroy the PG.

PR #380, opened 2025-08-31
Model: Revised context/cache size checks

DocShotgun:
The current context/cache size checks do not capture all cases that will trigger an assert in exllamav2/v3, specifically when the length of the prompt is within `max_seq_len` tokens, but there isn't sufficient allocated k/v cache to generate the requested number of `max_tokens`.


DocShotgun:
Superceded due to refactoring of the cache/sequence length handling.

PR #395, opened 2025-11-01
Add FLA and dependencies to Dockerfile

figuernd:
Qwen3-Next-80B requires `flash-linear-attention`, which is not included in the current Dockerfile. FLA is not prebuilt, so some build dependencies are also needed to support pip FLA install.

Adds FLA to Dockerfile and therefore Qwen3-Next support.


turboderp:
This has been superseded, so closing. Thanks though. <3

PR #407, opened 2026-01-18
Adaptive p sampling to exllamav3

Ph0rk0z:
I don't know if I did it a little stupidly inside sampler.py but it needs to be the last in the chain or else it fails. At least now it can be tested.


turboderp:
Oh, I'm sorry. Didn't even see there was already a PR. :|

Ph0rk0z:
Current implementation looks more clean to me.

PR #413, opened 2026-02-14
Full tool-calling support, inference abort fixes, XML parsing, OpenAI streaming compliance

devnen:
## What This PR Does

TabbyAPI's tool-calling system worked for simple cases but had a collection of bugs that surfaced the moment you pushed it harder — a different model family, a stricter client, or just hitting Stop at the wrong time. This PR fixes all of them in one pass, tested end-to-end against Kilo Code, Roo Code, and OpenCode with Qwen3-Coder-Next on a dual-GPU setup.

The changes fall into four areas:

- **Qwen3-Coder support** — Qwen3-Coder emits tool calls in XML, not JSON. TabbyAPI had no XML parser, so those tool calls were silently discarded. This PR adds full XML parsing and a new official Jinja template for the model family.
- **Streaming compliance** — Strict clients (OpenCode / Vercel AI SDK) were rejecting every streaming tool-call response due to missing fields and wrong chunk structure. The streaming path is now fully spec-compliant.
- **Stop actually stops** — Three independent bugs meant the model kept running on the GPU after a client disconnect. All three are fixed; GPU inference now aborts reliably on Stop.
- **Broader compatibility** — JSON parsing is hardened to handle the many ways real models deviate from the ideal output format, `tool_choice` is now respected, and a Jinja filter fix makes HuggingFace-native chat templates work out of the box.

> **Tested with:** ExLlamaV3 v0.0.22 · Qwen3-Coder-Next-exl3-4.0bpw · Kilo Code v5.6.0 · Roo Code · OpenCode (Vercel AI SDK). Windows 10 OS

Built against commit `41511f5` with ExLlamaV3 v0.0.22.

---

#### 1. XML Tool-Call Parsing

**Problem:** Qwen3-Coder models are trained to emit tool calls in XML format (`<function=name><parameter=key>value</parameter></function>`). TabbyAPI's tool-calling system only supported JSON via constrained generation, causing XML tool calls to be dumped as plain text with no `tool_calls` array in the response.

**Solution:**
- Added a regex-based XML parser (`from_xml()`) alongside the existing JSON path, with an `from_auto()` dispatcher that tries JSON → JSON-in-wrapper → XML in sequence.
- Added a `tool_call_format` metadata field to the Jinja template system, allowing templates to declare whether they expect `json`, `xml`, or `auto` format tool calls.
- Modified the two-pass generation flow: when XML mode is active, the second pass skips the JSON schema constraint and lets the model generate its natural XML output unconstrained.
- Added a fallback path that scans the `content` field for bare `<function=` patterns when the two-pass system doesn't trigger.
- Added pre-template argument conversion (`json.loads()` on string arguments) to prevent crashes in multi-turn tool conversations where clients send `arguments` as a JSON string but the Jinja template expects a dict.
- Added a new Jinja template (`templates/tool_calls/qwen3_coder.jinja`) based on the official Qwen3-Coder-Next `chat_template.jinja` with TabbyAPI metadata.

**Design notes:**
- Regex parsing (not XML parsing) is used because the format uses `<function=name>` with `=` in the tag name, which is invalid XML. This matches the approach taken by vLLM, llama.cpp, and the official Qwen parser.
- Type coercion uses `json.loads()` with string fallback, explicitly avoiding `eval()` (ref: CVE-2025-9141 in vLLM's parser).

#### 2. OpenAI Streaming Protocol Compliance

**Problem:** After adding XML parsing, non-streaming responses worked correctly, but strict clients (OpenCode / Vercel AI SDK) rejected streaming responses with `AI_TypeValidationError`. The SSE chunks were missing the required `index` field on tool-call deltas, emitting `role: "user"` instead of `"assistant"`, merging tool-call data with the finish signal, and leaking null fields.

**Solution:**
- Added `_build_tool_call_chunks()` implementing a two-chunk emission pattern: one chunk with complete tool-call data (`role: "assistant"`, `tool_calls` array with `index` values, `finish_reason: null`), followed by a separate finish chunk (`delta: {}`, `finish_reason: "tool_calls"`).
- Added `_serialize_stream_chunk()` for consistent serialization across all chunk types, using `exclude_none=True` while restoring semantically meaningful `finish_reason: null` on intermediate chunks.
- Restructured the streaming loop in `stream_generate_chat_completion()` to intercept tool-call generation results, parse them, and emit spec-compliant chunks before the normal chunk-building path.
- Removed tool-call handling from `_create_stream_chunk()` since it is now handled upstream.

#### 3. Pydantic v2 Union Coercion Fix

**Problem:** The `index` field added for streaming compliance was silently dropped during chunk construction. Pydantic v2's smart Union coercion converts dicts passed through `Union[ChatCompletionMessage, dict]` fields into `ChatCompletionMessage` instances, which in turn coerces nested tool-call dicts into `ToolCall` instances. With `extra='ignore'` (the default), any keys not declared on the model are silently discarded.

**Solution:**
- Added `index: Optional[int] = None` directly to the `ToolCall` model in `types/tools.py`. This ensures the field survives Pydantic's coercion rather than being treated as an extra field.
- Updated the tool-call ID factory to use the `call_` prefix convention (`call_<24-char hex>`), matching the format expected by some strict clients.

#### 4. Inference Abort Fixes

**Problem:** When a user presses "Stop" during streaming inference, TabbyAPI did not reliably stop generating tokens. The model continued running on the GPU after the client disconnected.

Three bugs were identified and fixed:

**Bug 1 — `gen_queue.get()` blocks disconnect detection:**
The consumer loop's `await gen_queue.get()` blocks indefinitely when the queue is empty (during prefill, between tokens). While blocked, the `disconnect_task.done()` check never re-executes.

*Fix:* Replaced the blocking `get()` with `asyncio.wait()` that races a queue get task against the disconnect task. Applied identically in both `stream_generate_completion` and `stream_generate_chat_completion`.

**Bug 2 — ExLlamaV3 job registered too late in `active_job_ids`:**
The `AsyncJob` was only assigned to `self.active_job_ids[request_id]` after the generation loop finished. During generation, the entry held `None`, which `wait_for_jobs()` skips.

*Fix:* Moved the assignment to immediately after `AsyncJob()` construction, before the generation loop.

**Bug 3 — `GeneratorExit` bypasses abort event:**
When `sse_starlette` crashes on a dropped TCP connection, it injects `GeneratorExit` into the async generator. The existing `except CancelledError` and `except Exception` handlers don't catch `GeneratorExit` (a `BaseException`), so `abort_event.set()` is never called and inference continues.

*Fix:* Moved `abort_event.set()` and `disconnect_task.cancel()` into a `finally` block, which executes on all exit paths including `GeneratorExit`.

---

### Files Changed

| File | Summary |
|------|---------|
| `backends/exllamav3/model.py` | Moved `active_job_ids` assignment before generation loop |
| `common/templating.py` | Added `tool_call_format` field to `TemplateMetadata`, validation |
| `endpoints/OAI/types/tools.py` | Added `index: Optional[int] = None` to `ToolCall`, `call_` prefix ID format |
| `endpoints/OAI/utils/chat_completion.py` | XML generation flow, streaming protocol compliance, `asyncio.wait`-based disconnect detection, `finally` block for abort |
| `endpoints/OAI/utils/completion.py` | `asyncio.wait`-based disconnect detection, `finally` block for abort |
| `endpoints/OAI/utils/tools.py` | XML parser (`from_xml`, `from_auto`, `parse`, `extract_content_and_tools`), think-block stripping, type coercion |
| `templates/tool_calls/qwen3_coder.jinja` | New file — official Qwen3-Coder-Next template with TabbyAPI metadata |

---

### Testing

Validated against:
- **OpenCode** (Vercel AI SDK) — strict Zod-based schema validation, previously rejected all streaming tool-call responses
- **Kilo Code** v5.6.0 — OpenAI-compatible VSCode extension
- **Roo Code** — OpenAI-compatible VSCode extension

Test scenarios: single and multiple tool calls, multi-turn tool conversations, mixed text + tool calls, streaming and non-streaming modes, client disconnect during inference.

### Environment

- ExLlamaV3 v0.0.22
- Qwen3-Coder-Next-exl3-4.0bpw (80B params, 3B activated, MoE)
- Dual GPU, 80K context window

---

**Edit:** Additional improvements after initial submission:

**5. Broader Model Compatibility**

- Added a `tojson` Jinja filter override so the model's built-in HuggingFace template works out of the box (the default sandboxed filter crashes on `tojson(ensure_ascii=False)` which Qwen3-Coder's template uses).
- Hardened JSON tool-call parsing to handle common model output variations: flat `{"name": ..., "arguments": ...}` dicts without the `function` wrapper, single objects instead of arrays, markdown-fenced JSON, and string-typed arguments. Previously only perfectly-formed OAI-shaped arrays were accepted.
- Generation chunks now include `token_ids` as a plain Python list, with robust handling for tensors and tuples from different ExLlamaV3 kernels.

**6. `tool_choice` Support**

Added support for the OpenAI `tool_choice` parameter: `"none"` skips tool generation entirely, `"required"` forces a tool-call pass even when the model doesn't emit the stop string, and named function choice (`{"type": "function", "function": {"name": "..."}}`) filters results to the specified function.

**7. Bug Fixes and Cleanup**

- Fixed a prompt mutation bug in `generate_tool_calls()` where the shared `prompt` variable was modified in-place during the loop. With `n > 1`, each iteration would stack previous generations' text onto the prompt. Now uses a per-iteration local variable.
- Added `tool_end` extraction to `TemplateMetadata` so template-provided metadata isn't silently discarded.
- Reduced debug logging verbosity in the XML parser — removed per-parameter and raw-text-dump log lines, kept summary-level logs and warnings.

**Additional files changed:**

| File | Summary |
|------|---------|
| `endpoints/OAI/types/chat_completion.py` | Added `tool_choice`, `parallel_tool_calls` fields |
| `endpoints/OAI/types/tools.py` | Added `NamedToolChoice`, `NamedToolFunction` models |

---

**Edit:** Thinking-model support added.

**8. Qwen3.5 Thinking-Model Template**

Added a new template, `templates/tool_calls/qwen3_thinking.jinja`, for Qwen3.5 and other Qwen thinking models. The existing `qwen3_coder.jinja` handles XML tool calls but has no awareness of `<think>...</think>` reasoning blocks, so thinking models using it produced malformed prompts on multi-turn conversations (reasoning text leaked into the next turn's content, and the generation prompt didn't open a `<think>` tag).

**What the new template does:**

- Splits assistant-history messages on `</think>`, extracting reasoning into `reasoning_content` and leaving only the post-think text as `content`. This matches the field convention already used by this PR's streaming path.
- Emits `<think>\n` at the end of the generation prompt so the model starts in reasoning mode, or `<think>\n\n</think>\n\n` when `enable_thinking=false` is passed as a template variable.
- Uses the same XML tool-call format as `qwen3_coder.jinja` (`<tool_call><function=...><parameter=...>`), so the existing XML parser handles output unchanged — no Python code changes required.
- Carries the same TabbyAPI metadata header (`tool_call_format = "xml"`, `tool_start`, `tool_end`, `stop_strings`) so it's a drop-in replacement: users point `prompt_template` at `qwen3_thinking` and everything else just works.

**Why this is safe as a pure template addition:**

The XML tool-call parser in this PR already strips `<think>...</think>` blocks (including unclosed ones from quantization degradation) before parsing, so reasoning output from a thinking model never interferes with tool-call extraction. The only thing missing was a template that correctly fed reasoning back into the model on subsequent turns — which this file provides.

Based on the official Qwen3.5 HuggingFace `chat_template.jinja`, with the TabbyAPI metadata header prepended. Validated against the workflow described in the [HuggingFace discussion for Qwen3.5-397B-A17B-exl3](https://huggingface.co/NeuroSenko/Qwen3.5-397B-A17B-exl3/discussions/1), which previously required a manual patch on top of this PR to get thinking + tool calls working together. With this template, no patch is needed — clone the branch, point `prompt_template` at `qwen3_thinking`, and tool calls + reasoning work end-to-end.

**Additional files changed:**

| File | Summary |
|------|---------|
| `templates/tool_calls/qwen3_thinking.jinja` | New file — Qwen3.5 thinking-aware template with XML tool calls |

### DIFF
--- a/backends/exllamav3/model.py
+++ b/backends/exllamav3/model.py
@@ -1014,8 +1014,21 @@ class ExllamaV3Container(BaseModelContainer):
                 if chunk:
                     chunk_tokens = result.get("token_ids", self.tokenizer.encode(chunk))
                     full_response += chunk
+
+                    # Extract token IDs as a plain list for downstream consumers
                     if isinstance(chunk_tokens, torch.Tensor):
+                        token_id_list = chunk_tokens.flatten().tolist()
                         generated_tokens += chunk_tokens.size(dim=0)
+                    elif isinstance(chunk_tokens, tuple):
+                        first = chunk_tokens[0]
+                        if isinstance(first, torch.Tensor):
+                            token_id_list = first.flatten().tolist()
+                        else:
+                            token_id_list = list(first)
+                        generated_tokens += len(token_id_list)
+                    else:
+                        token_id_list = list(chunk_tokens)
+                        generated_tokens += len(token_id_list)
 
                     # Increase penalty range to generated token amount
                     # TODO:
@@ -1025,6 +1038,7 @@ class ExllamaV3Container(BaseModelContainer):
                     generation = {
                         "request_id": request_id,
                         "text": chunk,
+                        "token_ids": token_id_list,
                         "prompt_tokens": context_len,
                         "generated_tokens": generated_tokens,
                         "offset": len(full_response),
--- a/backends/exllamav3/model.py
+++ b/backends/exllamav3/model.py
@@ -996,6 +996,7 @@ class ExllamaV3Container(BaseModelContainer):
             max_rq_tokens=self.max_rq_tokens,
             filters=grammar_handler.filters,
         )
+        self.active_job_ids[request_id] = job
 
         generated_tokens = 0
         full_response = ""
@@ -1044,8 +1045,6 @@ class ExllamaV3Container(BaseModelContainer):
 
                     yield finish_chunk
                     break
-            # Assign the active job to the request ID
-            self.active_job_ids[request_id] = job
 
         except asyncio.CancelledError:
             await job.cancel()
--- a/common/templating.py
+++ b/common/templating.py
@@ -24,12 +24,16 @@ class TemplateLoadError(Exception):
     pass
 
 
+VALID_TOOL_CALL_FORMATS = {"json", "xml", "auto"}
+
+
 @dataclass
 class TemplateMetadata:
     """Represents the parsed metadata from a template."""
 
     stop_strings: List[str] = field(default_factory=list)
     tool_start: Optional[str] = None
+    tool_call_format: str = "json"
 
 
 class PromptTemplate:
@@ -75,6 +79,18 @@ class PromptTemplate:
         if hasattr(template_module, "tool_start"):
             if isinstance(template_module.tool_start, str):
                 template_metadata.tool_start = template_module.tool_start
+
+        if hasattr(template_module, "tool_call_format"):
+            fmt = template_module.tool_call_format
+            if isinstance(fmt, str) and fmt in VALID_TOOL_CALL_FORMATS:
+                template_metadata.tool_call_format = fmt
+                logger.debug(f"Template tool_call_format: {fmt}")
+            else:
+                logger.warning(
+                    f"Invalid tool_call_format '{fmt}' in template, "
+                    f"defaulting to 'json'. "
+                    f"Valid values: {VALID_TOOL_CALL_FORMATS}"
+                )
 
         self.metadata = template_metadata
         return template_metadata
--- a/common/templating.py
+++ b/common/templating.py
@@ -12,6 +12,7 @@ from jinja2 import Template, TemplateError
 from jinja2.ext import loopcontrols
 from jinja2.sandbox import ImmutableSandboxedEnvironment
 from loguru import logger
+from markupsafe import Markup
 from packaging import version
 
 
@@ -33,6 +34,7 @@ class TemplateMetadata:
 
     stop_strings: List[str] = field(default_factory=list)
     tool_start: Optional[str] = None
+    tool_end: Optional[str] = None
     tool_call_format: str = "json"
 
 
@@ -49,6 +51,22 @@ class PromptTemplate:
         extensions=[loopcontrols],
     )
     metadata: Optional[TemplateMetadata] = None
+
+    @staticmethod
+    def _tojson_compat(value, indent=None, ensure_ascii=True):
+        """Compatibility JSON filter for chat templates.
+
+        Some model templates call ``tojson(ensure_ascii=False)`` while the
+        bundled Jinja filter may not accept that keyword in sandboxed mode.
+        """
+        return Markup(
+            json.dumps(
+                value,
+                indent=indent,
+                ensure_ascii=ensure_ascii,
+                separators=(",", ": "),
+            )
+        )
 
     async def extract_metadata(self, template_vars: dict):
         """
@@ -79,6 +97,10 @@ class PromptTemplate:
         if hasattr(template_module, "tool_start"):
             if isinstance(template_module.tool_start, str):
                 template_metadata.tool_start = template_module.tool_start
+
+        if hasattr(template_module, "tool_end"):
+            if isinstance(template_module.tool_end, str):
+                template_metadata.tool_end = template_module.tool_end
 
         if hasattr(template_module, "tool_call_format"):
             fmt = template_module.tool_call_format
@@ -123,6 +145,7 @@ class PromptTemplate:
 
         self.environment.globals["strftime_now"] = strftime_now
         self.environment.globals["raise_exception"] = raise_exception
+        self.environment.filters["tojson"] = self._tojson_compat
 
         return self.environment.from_string(template_str)
 
--- a/endpoints/OAI/types/chat_completion.py
+++ b/endpoints/OAI/types/chat_completion.py
@@ -4,7 +4,7 @@ from typing import Literal, Union, List, Optional, Dict
 from uuid import uuid4
 
 from endpoints.OAI.types.common import UsageStats, CommonCompletionRequest
-from endpoints.OAI.types.tools import ToolSpec, ToolCall
+from endpoints.OAI.types.tools import NamedToolChoice, ToolSpec, ToolCall
 
 
 class ChatCompletionLogprob(BaseModel):
@@ -71,6 +71,10 @@ class ChatCompletionRequest(CommonCompletionRequest):
 
     tools: Optional[List[ToolSpec]] = None
     functions: Optional[List[Dict]] = None
+    tool_choice: Optional[
+        Union[Literal["none", "auto", "required"], NamedToolChoice]
+    ] = None
+    parallel_tool_calls: Optional[bool] = True
 
     # Chat completions requests do not have a BOS token preference. Backend
     # respects the tokenization config for the individual model.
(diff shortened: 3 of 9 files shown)


mratsim (commented):
Given that this PR introduce a new jinja template. I think we should discuss (separate RFC issue?) what we should do with the `templates/` infrastructure in tabbyAPI. I don't think anyone actually uses it given that the one in this repo only contains alpaca and chatml, and the other repo at https://github.com/theroyallab/llm-prompt-templates as not been updated for 8 months and so does not support any of the best agentic model compatible with EXL3 (GLM-4.x, MiniMax-M2.x, Qwen3.5, ...)

devnen:
I see the main_tools branch picked up the tool parsing and abort fixes from this PR (commit 1794791). Nice to see it moving forward. The streaming path got a full rework there which looks like a better architecture than what I had.

devnen:
> Hi @devnen do you think you can add a keep alive that sends something so that Kilo/Cline/Roo etc don't disconnect while the response is still generating?
> 
> [Kilo-Org/kilocode#1681](https://github.com/Kilo-Org/kilocode/issues/1681)

The keepalive idea makes sense but it's really a separate feature from what this PR addresses. Might be worth
 opening a dedicated issue for it. SSE comment heartbeats during prefill would be straightforward to add.
 
Also, turboderp's rework on main_tools retained these abort fixes and may handle that scenario better with the new stream collector architecture. Worth testing that branch.

turboderp:
It's a confused standard. Very incomplete spec from OpenAI that leaves so many things undefined. But what's in the `main_tools` branch is tested fairly extensively against OpenCode, and I *believe* it should be robust overall, breaking the minimum number of clients.

I'll probably go over the actual parsing as well in the next few days. Currently it works well for Qwen3-Coder and Qwen3.5, but that's about it. I've been studying the various formats and no model seems to actually emit valid XML. Some do emit valid JSON that can be parsed pretty elegantly, but then you'll still need to normalize afterwards.

So really the only solution is to have a handwritten parser function per format, with not a lot of shared code between them. The nice thing though is that each function will be small and self-contained with a very simple contract of raw string in (with just tool calls), list of normalized tool calls out.

turboderp:
Oh, and keepalive makes sense (though do note that MoE prefill speed should be massively improved in recent versions of exl3, so maybe there's less need for this... 5 minutes of prefill seems like a lot.)

Tabby already receives updates from the backend every `max_chunk_size` tokens during prefill, so it could just stream an empty chunk in response to that. Once again the spec falls short, but I do believe most clients would tolerate a packet like `data: {"choices":[{"index":0,"delta":{}}]}` (?)

devnen:
Pushed qwen3_thinking.jinja for Qwen3.5 and other Qwen thinking models. qwen3_coder.jinja doesn't handle <think> blocks, so multi-turn breaks on thinking models.

Same XML tool-call format as qwen3_coder, same metadata header, drop-in. No Python changes. Point prompt_template at qwen3_thinking and thinking plus tool calls work end-to-end, no patch needed.

turboderp:
Closing this as stale. Tool calls seem to be working correctly with the `tool_calls` config option.

PR #432, opened 2026-07-13
Add ability to ban tokens on exl3 backend

ethanj801:
TabbyAPI exposes the fields `ban_eos_token` and  `banned_tokens` which can be used to allow for generation to continue indefinitely or to ban certain desired tokens. However this feature does nothing when using the exl3 backend. This PR resolves this.

These fields are present (or have equivalent features) across OAI compatible implementations (e.g. [textgen](https://github.com/oobabooga/textgen)). It also is present in TabbyAPI's own exl2 implementation.


ethanj801:
One could argue this sampler class should be present in exl3 rather than here (happy to move it there if desired).

ethanj801:
Going to move this into exl3 and then update tabby when exl3 updates

PR #436, opened 2026-07-25
fix(model): recognize symlink aliases as the loaded model during inline loading

### LINKED ISSUE #379: [BUG] Inline loading of models at Symlinks causes models to unload and reload
### OS

Linux

### GPU Library

CUDA 12.x

### Python version

3.12

### Describe the bug

I set `inline_model_loading: true` so I could load models dynamically from the `models` directory by name.

I saved the model from huggingface.co, such as `ArtusDev_Qwen_Qwen3-Coder-30B-A3B-Instruct-EXL3`, then I created a symlink named `coder` to that directory.

The goal was to be able to use "coder" as my model name for the API call while keeping a descriptive directory name on disk. This also lets me re-symlink to a different model without changing the clients making the calls.

However, I think it resolves the symlink, properly loads the model but thinks the model is called `ArtusDev_Qwen_Qwen3-Coder-30B-A3B-Instruct-EXL3` instead of `coder`, then when the next API call comes in for model `coder` it unloads the model, then immediately reloads it (its the same model).

### Reproduction steps

1) Set `inline_model_loading: true`
2) Save a model to the models directory. I used `ArtusDev_Qwen_Qwen3-Coder-30B-A3B-Instruct-EXL3`
3) `ln -s ./ArtusDev_Qwen_Qwen3-Coder-30B-A3B-Instruct-EXL3 coder`
4) Make an API call with the model name "coder"
5) The model will unload, then reload

### Expected behavior

The model would stay loaded and be used. 

### Logs

_No response_

### Additional context

This is super low priority

### Acknowledgements

Anai-Guo:
### Problem

Fixes #379.

With `inline_model_loading: true`, a model can be addressed by directory name in the request's `model` field. If that name is a **symlink** to the real model directory (the issue's example: a `coder` symlink pointing at `ArtusDev_Qwen_Qwen3-Coder-30B-A3B-Instruct-EXL3`), the model is unloaded and reloaded on *every* request even though it is already loaded.

### Cause

The "is this already loaded?" check compares the requested name against `container.model_dir.name`:

```python
loaded_model_name = container.model_dir.name
if loaded_model_name == model_path.name and container.loaded:
    ...  # already loaded, skip
```

But the container is created from `model_path.resolve()` (`common/model.py`), so `model_dir.name` is always the **real** directory name. A symlink alias (`coder`) never equals the resolved name (`ArtusDev_...-EXL3`), so the guard fails and the model is reloaded. The same bare-name comparison sits in the inline fast-path guard in `endpoints/OAI/utils/common_.py`.

### Fix

Compare **resolved paths** instead of bare directory names in both places, so a symlink that points at the currently-loaded model is recognized as the same model. Non-symlink paths resolve to themselves, so existing behavior is unchanged.

- `common/model.py` — `load_model_gen`: `container.model_dir.resolve() == model_path.resolve()`
- `endpoints/OAI/utils/common_.py` — `load_inline_model`: resolve the requested `model_dir / model_name` and compare against the loaded container's resolved dir.

### DIFF
--- a/common/model.py
+++ b/common/model.py
@@ -145,7 +145,12 @@ async def load_model_gen(model_path: pathlib.Path, **kwargs):
         if container and container.model:
             loaded_model_name = container.model_dir.name
 
-            if loaded_model_name == model_path.name and container.loaded:
+            # Compare resolved paths, not just directory names, so a symlink
+            # alias pointing at the already-loaded model is recognized as the
+            # same model instead of triggering a spurious unload/reload (#379).
+            already_loaded = container.model_dir.resolve() == model_path.resolve()
+
+            if already_loaded and container.loaded:
                 xlogger.info(f'Model "{loaded_model_name}" is already loaded')
 
                 # Emit a terminal progress event so API clients always
--- a/endpoints/OAI/utils/common_.py
+++ b/endpoints/OAI/utils/common_.py
@@ -62,9 +62,13 @@ def aggregate_usage_stats(usage_stats_list: list[UsageStats]) -> UsageStats:
 async def load_inline_model(model_name: str, request: Request):
     """Load a model from the data.model parameter"""
 
-    # Return if the model container already exists and the model is fully loaded
-    if model.container and model.container.model_dir.name == model_name and model.container.loaded:
-        return
+    # Return if the model container already exists and the model is fully loaded.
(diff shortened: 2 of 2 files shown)


turboderp:
Thanks. This was superseded in 7dfcafa, though, so I'm closing.

PR #437, opened 2026-07-27
Ignore HF's generation tags in jinja templates

gakada:
This allows to load e.g. [Laguna-XS-2.1](https://huggingface.co/turboderp/Laguna-XS-2.1-exl3), otherwise it will give an error:

```
jinja2.exceptions.TemplateSyntaxError: Encountered unknown tag 'generation'. Jinja was looking for the following tags: 'elif' or 'else' or 'endif'. The innermost block that needs to be closed is 'if'.
```

and chat completion won't work.

`IgnoreGenerationTags` extension is taken from [llama-cpp-python](https://github.com/abetlen/llama-cpp-python/blob/629bd1b333f60d24a01886c5f99019f4c7c3ea6c/llama_cpp/llama_chat_format.py#L196-L203).

### DIFF
--- a/common/templating.py
+++ b/common/templating.py
@@ -8,7 +8,7 @@ from datetime import datetime
 from importlib.metadata import version as package_version
 from typing import Optional
 from jinja2 import Template, TemplateError
-from jinja2.ext import loopcontrols
+from jinja2.ext import loopcontrols, Extension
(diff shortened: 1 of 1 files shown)


turboderp:
Thanks. This was superseded by 29680f4, though, with the addition of the Laguna tool parser.

PR #438, opened 2026-07-28
Prewarm native MTP batch shapes before serving requests

Dampish0:

Native Qwen3.5 MTP/recurrent CUDA graph shapes are captured lazily for each batch size. Warming a single request does not prepare later concurrency levels, so the first request at an unseen batch size can pay the full JIT and graph-capture cost.

In profiling, the first unseen batch-1 and batch-2 shapes generated at roughly 4 tokens/s. Once captured, the same model reached about 190 tokens/s at batch 1, 275-281 aggregate tokens/s at batch 2, and about 357 aggregate tokens/s at batch 4. The warmed scaling showed that steady-state kernels were healthy; the user-facing latency came from capture occurring after the API was already ready.

This adds an optional `model.mtp_warmup_batch_sizes` setting. During generator creation, TabbyAPI runs short, internal greedy jobs for each configured batch size before startup completes. This prepares both native MTP draft and target-verification shapes before real traffic arrives.

The implementation:

- is disabled by default;
- only runs for native MTP drafting;
- deduplicates and sorts configured batch sizes;
- rejects values outside `1..max_batch_size`;
- uses a fixed synthetic prompt and does not log or return generated text;
- documents the setting in `config_sample.yml`.

**Examples**

```yaml
model:
  max_batch_size: 4
  mtp_warmup_batch_sizes: [1, 2, 3, 4]
```

With batch sizes 1-4 configured, startup graph capture took approximately 49 seconds in the tested 150k-context Q8-cache/Q6-MTP configuration. After readiness, first external requests at concurrency levels 1-4 no longer encountered the capture stall.

**Additional context**

Validation performed:

- `python -m ruff check backends/exllamav3/model.py common/config_models.py`
- `python -m ruff format --check backends/exllamav3/model.py common/config_models.py`
- `python -m py_compile backends/exllamav3/model.py common/config_models.py`
- Pydantic config default and `[1, 2, 3, 4]` parsing checks
- Runtime checks for non-MTP no-op and out-of-range rejection
- Deterministic generation output matched the unpatched build by SHA-256
- Live concurrency 1-4 and internal API health checks completed without CUDA or runtime errors

### DIFF
--- a/backends/exllamav3/model.py
+++ b/backends/exllamav3/model.py
@@ -22,6 +22,7 @@ from exllamav3 import (
     Tokenizer,
 )
 from exllamav3.cache import CacheLayer_quant
+from exllamav3.generator.sampler import GreedySampler
 from backends.exllamav3.grammar import ExLlamaV3Grammar
 
 from backends.exllamav3.sampler import ExllamaV3SamplerBuilder
@@ -130,6 +131,7 @@ class ExllamaV3Container:
     max_batch_size: Optional[int] = None
     draft_num_tokens: Optional[int] = None
     ngram_match_min: int = 0
+    mtp_warmup_batch_sizes: List[int] = []
 
     def __init__(self):
         # Mutable state must be created per instance. A class-level default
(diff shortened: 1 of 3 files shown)


turboderp:
This took a while to get to because it's not really something that belongs in TabbyAPI. But there's an explicit warmup function in the backend now, exposed through `warmup: true` in Tabby's config. It runs an optimized pass that targets the specific kernel specializations for the loaded model under the current cache/generator settings.

PR #447, opened 2026-08-06
fix(deps): point the cu13 extra at ExLlamaV3 v1.3.0 wheels (#446)

### LINKED ISSUE #446: [BUG] pyproject.toml: cu13 extra still references older ExLlamaV3 wheels instead of v1.3.0
### OS

Linux

### GPU Library

CUDA 12.x

### Python version

3.12

### Describe the bug

The cu13 extra in pyproject.toml currently pins the ExLlamaV3 wheels from the v0.0.37 GitHub release, which installs ExLlamaV3 1.1.0.

Since ExLlamaV3 1.3.0 is now available for CUDA 13, could the cu13 dependency be updated to reference the newer wheels?

Updating the dependency would allow users installing via:

pip install -U .[cu13]

to receive the current ExLlamaV3 release without manually replacing the wheel after installation.

Unless there is a known compatibility issue with ExLlamaV3 1.3.0, this looks like a straightforward dependency version bump in pyproject.toml.

### Reproduction steps

pip install -U .[cu13]

### Expected behavior

Installs exllama v1.3.0

### Logs

_No response_

### Additional context

(I can post the PR that fixes this, but it's just 3 lines)

### Acknowledgements

Anai-Guo:
Fixes #446.

### Problem

The `cu13` extra still resolves the ExLlamaV3 wheels from the **v1.1.0** release:

```
exllamav3 @ .../releases/download/v1.1.0/exllamav3-1.1.0%2Bcu132.torch2.11.0-cp313-cp313-linux_x86_64.whl
```

so `pip install -U .[cu13]` installs ExLlamaV3 1.1.0, while the `cu12` extra
right above it already installs **1.3.0**. CUDA 13 users have to replace the
wheel by hand after every install to get the current release.

### Fix

Bump the six `cu13` wheel URLs from `v1.1.0` to `v1.3.0`. Nothing else changes —
the diff is 6 lines in one file, tag and filename only.

### Why this is safe

- **The wheels exist.** The `v1.3.0` release publishes exactly the six
  `cu132.torch2.11.0` variants the `cu13` extra references —
  `cp312`/`cp313`/`cp314` × `win_amd64`/`linux_x86_64` — the same matrix
  `v1.1.0` published. No marker in the extra is left without a wheel.
- **The build is `cu132.torch2.11.0` in both releases**, matching the
  `torch 2.11.0+cu130` pins in the same extra. The ABI target is unchanged.
- **1.3.0 is already exercised by this repo.** The `cu12` extra has been on
  `exllamav3` 1.3.0 since it was released (2026-07-31), so the backend code in
  `backends/exllamav3/` is already running against that version on CUDA 12.
  This only brings CUDA 13 to parity.
- `docker/Dockerfile.cu13` installs via `.[cu13]`, so it picks the bump up with
  no change of its own.

Credit to @Mario5Gray for reporting and for pinpointing it to the extra.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -102,12 +102,12 @@ cu13 = [
     "flash-linear-attention >= 0.5.0 ; python_version >= '3.12' and ((platform_system == 'Linux' and platform_machine == 'x86_64') or platform_system == 'Windows')",
 
     # Exl3
-    "exllamav3 @ https://github.com/turboderp-org/exllamav3/releases/download/v1.1.0/exllamav3-1.1.0%2Bcu132.torch2.11.0-cp314-cp314-win_amd64.whl ; platform_system == 'Windows' and python_version == '3.14'",
-    "exllamav3 @ https://github.com/turboderp-org/exllamav3/releases/download/v1.1.0/exllamav3-1.1.0%2Bcu132.torch2.11.0-cp313-cp313-win_amd64.whl ; platform_system == 'Windows' and python_version == '3.13'",
-    "exllamav3 @ https://github.com/turboderp-org/exllamav3/releases/download/v1.1.0/exllamav3-1.1.0%2Bcu132.torch2.11.0-cp312-cp312-win_amd64.whl ; platform_system == 'Windows' and python_version == '3.12'",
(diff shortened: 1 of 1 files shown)


turboderp:
:+1: Thanks

PR #452, opened 2026-08-19
API: Report prompt cache and speculative decoding counters in usage stats

DenysAshikhin:

The exllamav3 backend already computes prompt-cache and draft-token counters for
every request, but they never reach the API nor is there any way for a client to get these stats.
 `_create_response` builds a finish chunk with `cached_tokens` (`backends/exllamav3/model.py:1177`) and, whenever a draft model is active, `draft_accept` / `draft_reject` (`:1185-1191`); then `get_usage_stats` drops all three.

Today their only consumer is `common/gen_logging.py:93-94`, which prints them to the console. An API client has no way to see whether a prompt hit the cache or how well speculative decoding is performing, even though the numbers already
exist server-side. `backends/exllamav3/model.py:1184` carries a matching `# TODO: Add extended draft stats in backend`.

Prompt cache hit rate and draft acceptance are the two numbers that explain latency on an exllamav3 deployment. Surfacing them lets clients and proxies report effective cost, spot cache thrashing, and tune `draft_num_tokens` without scraping server logs.

Both use OpenAI's existing nested shape, so OpenAI-compatible clients pick them up with no changes:

- `usage.prompt_tokens_details.cached_tokens`
- `usage.completion_tokens_details.accepted_prediction_tokens` /
  `rejected_prediction_tokens`

**Examples**

Request (usage is gated on `stream_options.include_usage`):

```json
{
  "messages": [{"role": "user", "content": "<~2000 token prompt>"}],
  "max_tokens": 200,
  "stream": false,
  "stream_options": {"include_usage": true}
}

Sending the same prompt twice against a 27B exl3 model with MTP drafting
(draft_mode: mtp, draft_num_tokens: 3):

// call 1: cold
"usage": {
  "prompt_tokens": 1985,
  "prompt_tokens_details": { "cached_tokens": 0 },
  "prompt_time": 1.95,
  "completion_tokens": 73,
  "completion_tokens_details": {
    "accepted_prediction_tokens": 48,
    "rejected_prediction_tokens": 27
  },
  "total_tokens": 2058
}

// call 2: identical prompt, warm prefix cache
"usage": {
  "prompt_tokens": 1985,
  "prompt_tokens_details": { "cached_tokens": 1792 },
  "prompt_time": 0.17,
  "completion_tokens": 73,
  "completion_tokens_details": {
    "accepted_prediction_tokens": 48,
    "rejected_prediction_tokens": 27
  },
  "total_tokens": 2058
}

Additional context
- cached_tokens is rounded, since exllamav3 reports it fractional. - On the naming: OpenAI defines accepted_prediction_tokens /rejected_prediction_tokens for Predicted Outputs, where rejected tokens are still billed. Speculative decoding is the same accounting - tokens proposed ahead of time that the model confirmed or discarded - so the fields are reused
rather than inventing tabby-specific names. Happy to rename them inside completion_tokens_details if you'd prefer they stay distinct.
- ruff format --diff and ruff check both clean.

### DIFF
--- a/endpoints/OAI/types/common.py
+++ b/endpoints/OAI/types/common.py
@@ -6,13 +6,28 @@ from typing import Optional, Union
 from common.sampling import BaseSamplerRequest, get_default_sampler_value
 
 
+class PromptTokensDetails(BaseModel):
+    """OpenAI-style prompt token details (cache accounting)."""
+
+    cached_tokens: int = 0
+
+
+class CompletionTokensDetails(BaseModel):
+    """OpenAI-style completion token details (speculative decoding accounting)."""
+
+    accepted_prediction_tokens: int = 0
+    rejected_prediction_tokens: int = 0
+
+
 class UsageStats(BaseModel):
     """Represents usage stats."""
 
     prompt_tokens: int
+    prompt_tokens_details: Optional[PromptTokensDetails] = None
     prompt_time: Optional[float] = None
     prompt_tokens_per_sec: Optional[Union[float, str]] = None
     completion_tokens: int
+    completion_tokens_details: Optional[CompletionTokensDetails] = None
     completion_time: Optional[float] = None
     completion_tokens_per_sec: Optional[Union[float, str]] = None
     total_tokens: int
--- a/endpoints/OAI/utils/common_.py
+++ b/endpoints/OAI/utils/common_.py
@@ -1,6 +1,10 @@
 import pathlib
 from common import model
-from endpoints.OAI.types.common import UsageStats
+from endpoints.OAI.types.common import (
+    CompletionTokensDetails,
+    PromptTokensDetails,
+    UsageStats,
+)
 from common.tabby_config import config
 from common.auth import get_key_permission
 from common.logger import xlogger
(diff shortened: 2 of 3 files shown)


smallmj:
### Independent verification of the `cached_tokens` gap (exllamav3, 27B exl3 on single 3090)

While debugging why our harness couldn't read cache hit rates from a TabbyAPI instance serving an EXL3 quant, we hit exactly the problem this PR solves and confirmed it independently on `main` (b75fe27):

- `backends/exllamav3/model.py` computes `cached_tokens` on every finish chunk (the console log line `... 83% cached ...` comes from it)
- `get_usage_stats()` in `endpoints/OAI/utils/common_.py` drops it — `UsageStats` has no field for it, so `usage.prompt_tokens_details.cached_tokens` is never emitted

We patched it locally with the same shape proposed here (`prompt_tokens_details: {cached_tokens}` on `UsageStats`, populated from the finish chunk, plus a plain top-level `cached_tokens` for convenience) and it works as described: second identical request reports the warm-cache hit, clients pick it up with no changes.

Two small notes from our testing:

1. **Non-streaming `usage` is `null`** unless `stream_options.include_usage` is set (and streaming is used). Clients that don't set it see no usage object at all — worth documenting, since "I can't see cache stats" is often just this.
2. `exllamav3` reports `cached_tokens` as a fractional float; we round to int before emitting. The PR already handles this, just confirming it's necessary.

Happy to share our diff if useful, but it matches this PR's approach closely — no need to duplicate. Thanks for the fix, looking forward to it landing.

PR #458, opened 2026-08-29
API: Align with llama.cpp output so coding agents work out of the box

framp:
I want to connect pi-coding-agent with tabbyAPI, using https://github.com/freeyoung/pi-localllm-provider
I want to be able to autodetect settings like context size, vision / thinking support automatically.
Some tools send a developer role which is not supported.

Returning additional output and supporting other tools doesn't cost us much and provide value to end users.

**Examples**
Screenshot_20260829_170928

Without these changes the model is not detected / it's detected with hardcoded 32k context


turboderp:
I ended up implementing this a little differently, since, as far as I can tell, this would open config.json and stat every safetensors file in every model directory on every call. There are also no /api/tags or /api/show endpoint in llama_server (is that from Ollama?), and I'm really not sure about constants like `vocab_type=2` and `ftype="unknown"`.

I believe commit 815a3407 should cover what this is meant to do, though (pushing in a moment). Otherwise, please let me know.

PR #378, opened 2025-08-25
feat: Add XML tool call processing system for GLM-4.5 and Qwen3-Coder models

bold84:
## 🎯 Overview

This PR implements a comprehensive XML tool call processing system that enables both GLM-4.5 and Qwen3-coder models to work seamlessly with TabbyAPI's tool calling functionality. These models generate tool calls in different XML formats, but TabbyAPI expects OpenAI's JSON format. This implementation bridges that gap with a generic, extensible solution.

## 🚀 Key Features

- **Multi-Model Support**: Supports both GLM-4.5 and Qwen3-coder XML formats
- **Format Conversion**: Automatically converts XML tool calls to OpenAI JSON format
- **Extensible Architecture**: Generic XML processor system that can be extended for other XML-based models
- **Zero Breaking Changes**: Existing JSON tool calling functionality remains completely unchanged
- **Robust Error Handling**: Gracefully handles malformed XML and edge cases
- **Production Ready**: Comprehensive test coverage with 28 passing tests

### Core Implementation

- **`endpoints/OAI/utils/xml_tool_processors.py`**
  - `BaseXMLToolCallProcessor`: Abstract base class for extensible XML processing
  - `GLM45ToolCallProcessor`: GLM-4.5 specific XML parser
  - `Qwen3CoderToolCallProcessor`: Qwen3-coder specific nested XML parser (supports "qwen3-coder" only)
  - `XMLToolCallProcessorFactory`: Factory pattern for creating appropriate processors

- **`common/templating.py`**
  - Added XML-specific metadata fields: `tool_call_format`, `xml_processor_type`, `tool_start`, `tool_end`, `stop_strings`
  - Extended `extract_metadata()` to handle complete XML configuration
  - Maintains backward compatibility with existing templates

- **`endpoints/OAI/utils/tools.py`**
  - Added `from_xml()` method for XML-specific processing
  - Added `from_text()` method for automatic format detection and routing
  - Fixed import issues with `Tool` class

- **`templates/tool_calls/qwen3-coder-tabbyapi.jinja`**
  - Complete TabbyAPI-compatible template with full XML metadata
  - Includes stop_strings, tool_start, tool_end, tool_call_format, xml_processor_type

### Quality Assurance

- **`tests/test_xml_tool_calls.py`**
  - 28 comprehensive test cases covering all functionality
  - Tests for GLM-4.5 format: single/multiple tool calls, JSON arguments, error handling
[...]

PR #462, opened 2026-09-02
API: Stream partial tool_calls argument deltas for qwen3_coder formats

### LINKED ISSUE #460: Stream partial tool_calls argument deltas for coding agents (parity with vLLM/llama.cpp)
## Problem

When serving a tool-calling model through the OpenAI-compatible endpoint with `stream: true`, tool calls are buffered until generation finishes and delivered as a single `tool_calls` delta containing the complete arguments. Coding agents that render live tool generation (e.g. pi-coding-agent) show nothing while the model is producing a tool call, then receive the whole thing at once after a delay.

Repro (exllamav3 backend, `tool_format: qwen3_coder`, Qwen3.8-Flash-Next EXL3):

```
[  1.92s] TOOL_CALL delta: idx=0 id=call_... name=get_weather args='{"city": "Amsterdam"}'
[  1.92s] TOOL_CALL delta: idx=1 id=call_... name=get_weather args='{"city": "Tokyo"}'
[  1.92s] [DONE]
```

Both deltas arrive at the exact moment generation completes, with full arguments. The reasoning and content channels stream normally throughout, so only tool calls lag.

## Cause

This is the current design and it's documented. In `endpoints/OAI/utils/chat_completion.py` the streaming loop accumulates tool-channel text and only parses it on the final chunk:

```python
else:
    full_tool += sub
...
generation["delta_tool_calls"] = ""
if finish_reason and full_tool:
    generation["delta_tool_calls"] = _parse_tool_calls(full_tool, tool_format, request_id)
```

and the `_parse_tool_calls` docstring states:

> These are not choice indices; OAI enumerates the tool calls within each individual choice for the sake of streaming incomplete tool arg deltas, which we don't do here.

## Request

Emit incremental `tool_calls` deltas - function name first, then argument JSON fragments - so clients can render tool generation live. This matches vLLM (`--tool-call-parser qwen3_coder`) and llama.cpp, both of which stream partial arguments.

I understand why it's non-trivial: the pseudo-XML format needs to be converted to a valid JSON `arguments` string incrementally, and whitespace stripping + `coerce_param_value` type coercion are only knowable at the closing `parameter` tag. A scheme that keeps the accumulated result byte-identical to the current end-parse:

- On the opening `function=NAME` tag: emit `{index, id, function: {name}}`
- On each `parameter=k` tag: emit `{"k": "` (or `,"k": "` for subsequent params)
- String values: stream live with leading/trailing whitespace holdback (the same trick `TagStreamParser` already uses for partial tags)
- JSON-coercible values (arrays/objects/numbers): hold until the closing `parameter` tag and emit the coerced literal
- Keep the current authoritative `_parse_tool_calls` on the final chunk as the source of truth

Happy to contribute a reference implementation for `qwen3_coder` if there's interest - running exllamav3 + Qwen3.8-Flash-Next EXL3 in a pi coding-agent session, and this is the main UX gap vs vLLM for that workload. Saw #458 is also targeting pi integration.

tengotengo:
## What

Streaming responses with a qwen3_coder-family format (`qwen3_coder`, `qwen3_5`, `step3_5`,
`step3_7`) now report tool calls **while the model generates them**: `tool_calls` deltas with
partial `arguments` fragments, instead of one complete delta after generation finishes.
Closes #460.

## How

`QwenToolCallDeltaStreamer` consumes the TOOL-channel text that `TagStreamParser` already produces
and turns the pseudo-XML structure into OAI deltas: an open frame (`index`, `id`, `type`, `name`,
`arguments: "{"`), argument fragments as parameters close, `"}"` when the function closes.

- String values stream live as soon as they provably cannot parse as a JSON literal, so
  `coerce_param_value()` cannot reinterpret them later. Values it *may* convert (objects, arrays,
  numbers, booleans, null, quoted strings) are emitted whole when their parameter closes.
- Malformed tag sequences are read the way the end-of-stream regexes read them: stray close tags
  and parameters outside a function block are dropped, a nested function open belongs to the outer
  call. The streamer never raises into the collector.
- The authoritative end-of-stream parse stays the fallback whenever the streamer produced nothing.
- `verify()` compares the streamed assembly against the end-of-stream parse at the end of every
  request. It is log-only, because fragments cannot be retracted.

Only the qwen3_coder family is wired up; harmony, muse_glimmer and the rest keep the current path.

## Guarantee

For well-formed calls the concatenated `arguments` fragments per index are byte-identical to what
the end-of-stream parser produces, under random, one-character and whole-text chunkings — a client
that assembles deltas sees exactly what a non-streaming request returns.

Three inputs diverge on purpose, and are documented in the module docstring and the docs page:

1. **A call generation was cut off inside.** The regex needs a closing tag and drops the whole
   call, while the client already holds the partial arguments; `verify()` names that cause
   explicitly. Previously the same request returned `finish_reason: "tool_calls"` with an empty
   `tool_calls` list.
2. **Duplicate parameter keys.** Both are streamed (they parse to the same object); the parser
   keeps the last one.
3. **A value containing a literal parameter close tag.** Both sides end the value at the first
   occurrence but read the remainder differently.

## Testing

- Unit: `python -m unittest tests.test_toolcall_stream tests.test_tool_delta_streaming` (61 tests).
  The whole repo suite — 182 tests — passes on Python 3.10 and 3.11.
- `ruff check` and `ruff format --diff` clean.
- Live: `python tests/req_tool_delta.py` runs 16 contract checks against a running server
  (Qwen3.8-Flash-Next, ExLlamaV3): frame shape and ordering, streamed vs non-streaming equality,
  `tool_choice` none/required/named, reasoning-then-tool and content-then-tool ordering, client
  disconnect mid-call, `n>1`, `logprobs`, non-streaming regression, truncated calls. Normal traffic
  produces no `verify()` errors in the server log.
- The tests were checked against deliberate breakage of each promised behaviour (24 mutations:
  21 caught, 3 provably equivalent), so the suite fails when the guarantee breaks. One caveat
  found that way: fixtures must take their wrapper tags from the format itself, otherwise the text
  never reaches the tool channel and every comparison silently passes on empty input.

## Also included

One-line fix in `tests/_common.py`: `tool_calls` is a list of per-choice dicts, so the old
membership test was never true and the entry of a call was recreated on every delta, keeping only
the last argument fragment. Invisible while calls arrived as a single delta; visible as soon as
arguments are streamed in fragments.

## Relationship to #378

#378 adds an XML tool-call *parsing* path (it predates the current `toolcall_formats` package) and
touches the same two files for that purpose. This change assumes the existing format parsers and
only changes how their input is streamed, so the two would need a rebase but address different
problems.


turboderp:
I see a "what" and a "how", but I'm not entirely clear on the "why". Do clients benefit from streaming tool calls at all?

fubyo:
> I see a "what" and a "how", but I'm not entirely clear on the "why". Do clients benefit from streaming tool calls at all?

I like this.

It's more of a user experience improvement. Imagine the tool call writing a huge file. Now, you don't know what's happening before the tool call is complete, which might take a while.

Additionally, with this you could cancel a long tool call midway, if the user notices that the agent goes in the wrong direction.

acasar:
+1 for this! This has been one of the things that's been bugging me the most with tabbyAPI compared to llama.cpp.

When an agent is writing a long file, there is no progress visible in the harness, sometimes for minutes. In llama.cpp I can always follow the progress of a long tool call and even abort it if it's not going in the right direction. It would be amazing to have the same in tabbyAPI.

PR #480, opened 2026-09-17
Add Spark X2.5 tool call format (spark2_5)

dfein38347g:
## Summary

Adds a tool-call parser for **Spark X2.5** (and the X2.5-9B family) — bar-pipe tagged blocks with a bare function name and `<arg_key>`/`<arg_value>` argument pairs.

Raw format as emitted by the model:

```
<|tool_call|>
add<arg_key>a</arg_key><arg_value>2</arg_value><arg_key>b</arg_key><arg_value>3</arg_value>
<|/tool_call|>
```

Select it in the yml with `tool_format: spark2_5` (registry entry added to `ALL_TOOLCALL_FORMATS`).

## Changes

- `endpoints/OAI/utils/toolcall_formats/spark2_5.py` — new format module (`TOOLCALL_START`/`TOOLCALL_END` + `parse_toolcalls`, same module conventions as the other formats; values go through `coerce_param_value`).
- `endpoints/OAI/utils/tools.py` — import + `"spark2_5": spark2_5` registry entry.

## Behavior

- A call is delimited by the bar-pipe tags; the function name is the word immediately after the opening tag, arguments are the `<arg_key>/<arg_value>` pairs that follow.
- Multiple blocks in one generation are all parsed; an unterminated trailing block (stream end) is parsed from what follows the opening tag.
- No match => empty list, never raises (matches the other formats).

## Testing

Verified against the real 1.5.0 source in a live container serving Spark-X2.5-4B (exllamav3 1.5.0): 5-case in-container run — canonical full block, unterminated block, two consecutive blocks, pipe-less degenerate emission (correctly NOT matched — see note in the module docstring), and string-vs-number value coercion. All pass, including the registry and `get_toolcall_tags("spark2_5")` checks.

## Known caveat (model-side, not parser-side)

Small Spark variants occasionally emit the block without the pipe characters (a bare `
...
`). The stream router keys off the tagged form, so those runs pass through as content; a client-side fallback parser is the appropriate place for that, and it is out of scope here.


turboderp:
Thank you, but as far as I can tell, this model uses exactly the GLM4.5 tool format. There's no `<|tool_call|>` token in the vocabulary either, so I don't think the format defined here is correct.

I've added spark2_5 as an alias to glm4_5 for now. It seems to work. (?)

